PR #2, opened 2021-11-10
Integrate Verifiable Presentation Registry (preview version)

moritzkirstein:
Adds the Verifiable Presentation Registry (see: https://github.com/deltaDAO/vp-registry)

This feature is under heavy development and additional implementations will be made in the near future.

Currently supports simple registration & updating of a JSON file using the [GaiaxCredential template](https://github.com/walt-id/waltid-ssikit-vclib/blob/master/src/main/kotlin/id/walt/vclib/vclist/GaiaxCredential.kt) provided in the SSIKIT by @walt-id

Verification status is displayed on Asset Detail pages if the registry finds a valid VP at the given URL


fr-3deric (commented):
🚀

src/components/atoms/VerifiedPublisher.module.css
```diff
@@ -0,0 +1,29 @@
...
+}
+
+.verifiedBadge svg path {
+  fill: var(--brand-white) !important;
+}
+
+.loader span,
+.verifiedBadge span {
+  padding: 0 calc(var(--spacer) / 4);
+}
```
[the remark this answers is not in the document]
moritzkirstein: no particular order so far I think. I like the idea of ordering the CSS tho, we can start by adding this to our adjustments.

src/components/pages/Profile/History/Verify.tsx
```diff
@@ -0,0 +1,118 @@
...
+interface VerifyFormData {
+  vp: string | File[]
+}
+
+const initialValues = {
+  vp: ''
+}
+
+const validationSchema = Yup.object().shape({
+  // ---- required fields ----
```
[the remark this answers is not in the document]
moritzkirstein: Yes, I looked at how this was done on other forms. `required()` is simply here to adjust the error message on empty fields.

src/components/atoms/VerifiedPublisher.tsx
```diff
@@ -0,0 +1,44 @@
...
+      axios
+        .get(`${vpRegistryUri}/vp/${address}/verify`)
+        .then((response) => {
+          Logger.debug('[Verification] publisher verification:', response.data)
+          setVerified(response.data?.data?.verified)
+        })
+        .finally(() => {
+          setLoading(false)
+        })
+    }
```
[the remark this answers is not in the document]
moritzkirstein: Since I implemented async/await in the other places in this PR, I will revert to that and restructure the code here.

It could be a good idea to inform the user in case of server errors / unreachable API, since publishers may be verified, but the API is not responding for example.

Will have to think on some solutions to display that, but this is probably a new issue rather than something for this PR

src/@types/Verification.d.ts
```diff
@@ -0,0 +1,27 @@
+export interface RegisterVPPayload {
+  signature: string
+  hashedMessage: strng
```
oceanByte, line 3: issue(blocking): fix to `string`

src/components/atoms/VerifiedPublisher.module.css
```diff
@@ -0,0 +1,31 @@
...
+  font-weight: var(--font-weight-bold);
+}
+
+.verifiedBadge svg {
+  display: inline-block;
+  height: 100%;
+}
+
+.verifiedBadge svg path {
+  fill: var(--brand-white) !important;
```
oceanByte, line 25: nitpick(blocking): we should avoid this pattern as suggested by MDN: https://developer.mozilla.org/en-US/docs/MDN/Guidelines/Code_guidelines/CSS#dont_use_!important
is there a different way to achieve this?

src/components/pages/Profile/History/VerifyForm.tsx
```diff
@@ -0,0 +1,125 @@
...
+    setTouched,
+    status,
+    validateField
+  } = useFormikContext()
+
+  useEffect(() => {
+    setErrors({})
+    setTouched({})
+
+    // setSubmitting(false)
```
oceanByte, line 61: nitpick(blocking): We should remove comments as they confuse other readers.

src/components/pages/Profile/History/VerifyForm.tsx
```diff
@@ -0,0 +1,125 @@
...
+  }, [setErrors, setTouched])
+
+  const { accountId } = useWeb3()
+  const [addOrUpdate, setAddOrUpdate] = useState('Add')
+  useEffect(() => {
+    const initFileData = async () => {
+      try {
+        const response: RegistryApiResponse<VpDataBody> = await axios.get(
+          `${vpRegistryUri}/vp/${accountId}`
+        )
```
oceanByte, line 71: issue(non-blocking): let's go with this approach for now but in the long run we should avoid this kind of call and use a facade pattern here. This allows us to switch implementation from Axios to any other library if we should ever decide to change it. Take this as an architecture improvement hint for our future projects.

oceanByte (changes requested):

src/@types/Verification.d.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export interface SignatureMessageBody {
+  message: string
+}
+
+export interface VpDataBody {
+  _id: string
+  address: string
+  created: number
+  updated: number
```
oceanByte, line 24: question(blocking): I think we do not use these properties anymore? Let's remove them if they are not needed.

oceanByte (changes requested):

src/@types/Verification.d.ts
moritzkirstein: agreed

oceanByte (approved):

PR #11, opened 2021-11-30
Header refactor

### LINKED ISSUE #10: [Enhancement]: Refactor Code (styles) to use less "magic numbers"
thought(non-blocking): I prefer to work with multiple base values. When dealing with pixel-perfect positioning, we can add additional base values. But in my experience, it can be very difficult to maintain magic numbers across multiple projects. Using base values also gives non-designers a grid to follow when designing new elements. 

We need to discuss this in-depth after the hackathon @moritzkirstein Let's merge this PR.

_Originally posted by @oceanByte in https://github.com/deltaDAO/mvg-portal/pull/8#pullrequestreview-817784455_

LucaMilanese90:
Fixes #10 

## Proposed Changes

  - header refactor to avoid using absolute positioning

### DIFF
--- a/src/components/organisms/HomeIntro.tsx
+++ b/src/components/organisms/HomeIntro.tsx
@@ -1,14 +1,11 @@
 import React, { ReactElement } from 'react'
 import { useStaticQuery, graphql } from 'gatsby'
-import Container from '../atoms/Container'
 import styles from './HomeIntro.module.css'
-import { ReactComponent as Play } from '../../images/play.svg'
 import Markdown from '../atoms/Markdown'
 import {
   SortDirectionOptions,
   SortTermOptions
 } from '../../models/SortAndFilters'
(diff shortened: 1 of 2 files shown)


src/components/organisms/HomeIntro.tsx
```diff
@@ -58,41 +48,34 @@ interface HomeIntroData {
...
+            <div className={styles.circle}>
+              <a
+                href="https://academy.delta-dao.com/"
+                target="_blank"
+                rel="noopener noreferrer"
+              >
+                <h2 className={styles.circleText}>start now</h2>
+              </a>
+            </div>
           </div>
```
oceanByte, line 71: thought(non-blocking): looks like something that we could extract to a component.

oceanByte (approved):

PR #26, opened 2021-12-02
add macOS installation instructions

oceanByte:
Add an explanation of how to install node dependencies for macOS running on ARM


oceanByte:
The installation also works for macOS ARM without any custom command nowadays. PR closed.

PR #58, opened 2022-02-21
Feat: add third party portals section to home page

### LINKED ISSUE #52: [Feature] 3rd Party Portal Section
## Motivation / Problem
We have already demonstrators with various use-cases live and in use. These demos can and should be linked from the MVG landing page. We need to prepare a basic design to integrate 2 (or more in future) portals for now.

## Solution
- [x] prepare design on figma
- [x] implement the design

LucaMilanese90:
Closes #52 

## Proposed Changes

  - new third party portals section in home page


moritzkirstein:
Can you quickly bump this to get the preview build triggered @LucaMilanese90 ?

src/components/pages/Home.tsx
```diff
@@ -135,6 +168,14 @@ export default function HomePage(): ReactElement {
           <HomeContent />
         </section>
         <Container>
+          <div className={styles.section}>
+            <h3>Featured Portals</h3>
+            <div className={styles.portals}>
+              {portals?.map((portal, i) => (
+                <PortalTeaser {...portal} key={i} />
```
moritzkirstein, line 175: opinion: maybe we can start with a convention here to set keys. It would make reviewing a little easier and also follow best practices anyways if they are more unique (like @oceanByte mentioned before)

Maybe we can have the component / context integrated into the key: something like `portal-teaser-${i}` in this case. In cases of `<li>` elements, we could think about naming accordingly to content. In the case of a portal list: `portal-link-${i}` or similar.

I'd like some opinions on this and settle on something to go forward.

moritzkirstein (changes requested):

src/components/pages/Home.tsx
LucaMilanese90: I don't see how adding a fix string before the index would make any difference (apart from maybe readability), am I missing something? 
I went with index because there are only two static items (fetching the content from a json file) and they are not being manipulated in any way, just displayed, so it should work without problems. 
But I do agree that having a convention for this cases would be better as I find myself all the times wondering on the approach to follow.

moritzkirstein: As I recall it the main problem for dynamic content and keys was that those may not be unique over a whole page of components. If however we find a way to always specify the context within the key we should not have duplicate keys distributed over a page anymore, correct? At least if we make sure the naming convention considers this.
Maybe I am confusing stuff here, but that was my idea behind the example.

I would argue to then always use the convention, independent from if it is static or dynamic content - this is mainly for consistency and an easier reviewing process.

LucaMilanese90: This case falls into the "exception from the rule", should I still update it? And what about all the other components in the repo? Should I then update those as well to include a nanoid generated id?

oceanByte: No need to update this case if it is covered by the "exception from the rule". 

In general, I do not think that refactoring our entire codebase would add too much value here. We should just follow this agreement when touching old code anyways (refactoring) or when we introduce new features.

moritzkirstein: Fully agree on not to refactor the entire codebase right now. We should just consider this for future additions to the codebase. Thanks for sharing the reference again!

I believe using something like `nanoid` as a default from now on will however ease the review process and also make sure we always follow best practice here and can't introduce any errors by accident.

LucaMilanese90: nanoid comes with its own trade offs though, it can't be used directly generate the key inside a .map() function, otherwise it will change all the times and it will create more problems than it solves. All the arrays have to be integrated before use to add the id value on component load.
It could maybe even be useful to write a utility function to do it automatically.

moritzkirstein: Yes agreed. It is probably worth moving this out of this particular PR. But opening an issue to track this and start introducing a solution for future additions.

moritzkirstein (approved):

PR #59, opened 2022-02-21
Enhancement: new footer design

### LINKED ISSUE #51: [Enhancement] New Footer design
## Motivation / Problem
The footer currently is really small and cannot handle more information as is. We want to introduce some new content to the footer which would need more space.

## Solution 
- [x] prepare a footer design on figma
- [x] implement the new footer design

LucaMilanese90:
Closes #51 

## Proposed Changes

  - new footer design


moritzkirstein:
Nice!! Thank you for the quick PR.

I have one remark from browsing the preview:
grafik

Would like to have the headings stryled just a little more different from the content. Let's maybe try a bold font and secondary color.

src/components/organisms/Footer/Links.module.css
```diff
@@ -0,0 +1,46 @@
...
+  width: 100%;
+}
+
+.section {
+  text-align: left;
+}
+
+.title {
+  margin-bottom: calc(var(--spacer) / 2);
+  text-transform: uppercase;
```
moritzkirstein, line 15: suggestion: maybe we can try `--font-weight-bold` and `--color-secondary` on the titles

src/components/organisms/Footer/Links.tsx
```diff
@@ -0,0 +1,51 @@
...
+            {section.links.map((e, i) => (
+              <Link key={`${e.name}-${i}`} to={e.link}>
+                {e.name}
+              </Link>
+            ))}
+          </div>
+        </div>
+      ))}
+      <div className={styles.section}>
+        <p className={styles.title}>PRIVACY</p>
```
moritzkirstein, line 31: suggestion: could actually also be moved into the content `.json` to allow chaning the title more easily

src/components/organisms/Footer/Links.module.css
```diff
@@ -0,0 +1,46 @@
+.container {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: space-between;
```
moritzkirstein, line 4: grafik

suggestion/opinion(non-blocking): I think I'd like a `flex-start` styling more on mobile as well. It can look a little weird on tablet sized screens, but I prefer it over the stretched variant we get with `space-between`

src/components/organisms/Footer/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+export default function Footer(): ReactElement {
+  // TODO: replace placeholder footer content
+  const { footer } = useSiteMetadata()
+  const { title, subtitle, copyright } = footer
+
+  return (
+    <footer className={styles.footer}>
+      <div className={styles.container}>
+        <div className={styles.main}>
+          <FooterLogo />
```
moritzkirstein, line 17: issue(blocking): Let's use a more descriptive name here. Maybe something like `hashtag.svg` and simply import as `Hashtag`

moritzkirstein (changes requested):

src/components/organisms/Footer/Links.tsx
LucaMilanese90: Yes, it's like this because the whole privacy section is a little bit weirdly built with all these exceptions but I can add this to the json.

moritzkirstein (approved):

PR #62, opened 2022-02-23
Feat: quotes section

### LINKED ISSUE #53: [Feature] Quote Section
## Motivation / Problem
We want to display some quotes regarding the MVG on the portal landing page.

## Solution
- [x] prepare design on figma
- [x] implement design

LucaMilanese90:
Closes #53

## Proposed Changes

  - add quotes section to home page


src/components/organisms/SectionQuotes.tsx
```diff
@@ -0,0 +1,134 @@
...
+      childContentJson {
+        title
+        quotes {
+          name
+          profilePicture
+          quote
+        }
+      }
+    }
+    profilePictures: allFile(
```
moritzkirstein, line 22: issue(blocking): One problem with this approach is that the name of the file as well as the `profilePicture` in the `quotes.json` always need to be synced - without a clear link for any *non-coding contributor*. Meaning the link is made here in code, and not in the `.json` file.

I think in the siteMetadata a similar issue was solved pretty smoothly. Have a look at the `siteImage` in `site.json`. This is then accessed in the `hooks/useSiteMetadata.ts` hook.
Basically, given a valid filepath, the string is automatically mapped to a file (image in this case) and the necessary `childImageSharp` props can be used as below.

Would be a nice solution, if this can be applied here, too.

src/components/organisms/SectionQuotes.tsx
```diff
@@ -0,0 +1,134 @@
...
+  currentQuoteRef.current = currentQuote
+
+  const nextQuoteTimer = useRef(null)
+
+  useEffect(() => {
+    if (!data) return
+    setFullQuotes(
+      data.file.childContentJson.quotes.map((quote) => {
+        const picture = data.profilePictures.edges.find(
+          (e) => e.node.name === quote.profilePicture
```
moritzkirstein, line 85: issue(blocking): See comment above. Maybe this link made in code can be removed entirely.

src/components/organisms/SectionQuotes.tsx
```diff
@@ -0,0 +1,134 @@
...
+  id: string
+  name: string
+  profilePicture: string
+  quote: string
+}
+
+export default function SectionQuotes(): ReactElement {
+  const data: QuotesContentData = useStaticQuery(query)
+  const { title } = data.file.childContentJson
+  const [fullQuotes, setFullQuotes] = useState<Quote[]>([])
```
moritzkirstein, line 72: nitpick/opinion(non-blocking): I think just `[quotes, setQuotes]` would be enough.

moritzkirstein (changes requested):
I love the puppies & kittens 😍 😄 
A quick first review is in place. Nice work, I really like the design ✨ 👍

src/components/organisms/SectionQuotes.tsx
```diff
@@ -0,0 +1,110 @@
...
+        <h3 className={styles.sectionTitle}>{title}</h3>
+        <div className={styles.images}>
+          {quotes.map((e, i) => (
+            <img
+              key={e.id}
+              className={cx({
+                profilePicture: true,
+                active: e.id === quotes?.[currentQuote].id
+              })}
+              alt="profile-picture"
```
moritzkirstein, line 97: suggestion(blocking): We should probably also add the name of the quotee in the alt text, e.g. `Profile picture ${e.name}`

moritzkirstein (changes requested):

moritzkirstein (approved):

PR #64, opened 2022-02-24
Enhancement: remove video section

### LINKED ISSUE #63: [Enhancement] Remove Video Section
## Motivation / Problem
We need to remove the YT video from the landing page due to privacy implications.

## Solution 
For now, we should simply remove the video player and use a placeholder image. We can use the CtD section image.

## Notes
Since this is the only place where we actually displayed any video, we should also remove all related dependencies.

LucaMilanese90:
Fixes #63

## Proposed Changes

  - remove video player
  - extract modal image component structure


src/components/organisms/HomeContent.tsx
```diff
@@ -51,7 +50,7 @@ interface HomeContentData {
           body: string
           cta: string
           ctaTo: string
-          interactivity: string
+          interactivity: 'image'
```
moritzkirstein, line 53: suggestion(blocking): It seems like we now only support `'image'` "interactivity". I believe this justifies a refactoring of the component(s) here. We can basically utilize the same image loading via graphql as in `siteMetadata` (meaning relative path inside the content `.json`)
We could then get rid of the `interactiveComponents` object and handle rendering again in the `paragraphs.map()` completely.

moritzkirstein (changes requested):

src/components/organisms/HomeContent.tsx
```diff
@@ -109,12 +80,20 @@ export default function HomeContent(): ReactElement {
               }
             >
               <div className={styles.interactivity}>
-                {interactiveComponents[paragraph.interactivity]}
+                <InteractiveModalImage
+                  src={paragraph.image.childImageSharp.original.src}
+                  alt="Compute-to-Data and its benefits"
```
moritzkirstein, line 85: issue/suggestion(blocking): we should make sure to always use an alt text associated with the images. I suggest to simpl use the title of the section here for now.

moritzkirstein (changes requested):

PR #70, opened 2022-02-28
Fix: adjust url input & publish transformation

### LINKED ISSUE #68: [Enhancement] XSS File Input
## Motivation / Problem
Have a look at https://github.com/oceanprotocol/market/commit/9f2cbaafc1fbae74303ef8b478677a9a46d0d385

## Solution 
Merge/Implement XSS fix changes into the demonstrator

moritzkirstein:
Fixes #68 

## Proposed Changes

  - adjust `URLInput` component
  - adjust metadata transformation

### DIFF
--- a/src/components/molecules/FormFields/URLInput/Input.tsx
+++ b/src/components/molecules/FormFields/URLInput/Input.tsx
@@ -1,9 +1,10 @@
-import React, { ReactElement } from 'react'
+import React, { ReactElement, useEffect, useState } from 'react'
 import Button from '../../../atoms/Button'
 import { FieldInputProps, useField } from 'formik'
 import Loader from '../../../atoms/Loader'
 import styles from './Input.module.css'
 import InputGroup from '../../../atoms/Input/InputGroup'
+import isUrl from 'is-url-superb'
 
 export default function URLInput({
   submitText,
@@ -16,6 +17,20 @@ export default function URLInput({
   isLoading: boolean
 }): ReactElement {
   const [field, meta] = useField(props as FieldInputProps<any>)
+
+  const [buttonDisabled, setButtonDisabled] = useState(true)
+
+  useEffect(() => {
+    if (!field?.value) return
+
+    setButtonDisabled(
+      !field?.value ||
+        field.value === '' ||
+        !isUrl(field.value) ||
+        field.value.includes('javascript:') ||
+        meta?.error
+    )
+  }, [field?.value, meta?.error])
 
   return (
     <InputGroup>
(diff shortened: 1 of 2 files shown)


src/utils/metadata.ts
```diff
@@ -99,6 +99,22 @@ function getAlgorithmFileExtension(fileUrl: string): string {
...
+  return (
+    typeof files !== 'string' &&
+    files?.length && [
+      {
+        ...(files as File[])[0],
+        url: sanitizeUrl((files as File[])[0].url)
+      }
+    ]
+  )
+}
```
oceanByte, line 116: suggestion(non-blocking): This is just a simple check and unnecessarily difficult to read. Extracting the object in the array to a variable might improve readability here.

src/utils/metadata.ts
```diff
@@ -113,21 +129,24 @@ export function transformPublishFormToMetadata(
 ): MetadataMarket {
   const currentTime = toStringNoMS(new Date())
 
+  const transformedLinks = typeof links !== 'string' &&
```
oceanByte, line 132: nitpick(non-blocking): It is basically the same check as above. Could this be generalized into one function? Or could at least both use the same style (both being a function).

src/components/molecules/FormFields/URLInput/Input.tsx
```diff
@@ -17,6 +18,20 @@ export default function URLInput({
...
+
+  useEffect(() => {
+    if (!field?.value) return
+
+    setButtonDisabled(
+      !field?.value ||
+        field.value === '' ||
+        !isUrl(field.value) ||
+        field.value.includes('javascript:') ||
+        meta?.error
```
oceanByte, line 31: suggestion(non-blocking): When having so many conditions, it might be beneficial to extract this into a separate function to improve readability. As we are dealing with an XSS attack here, the function could be named accordingly or have a comment attached to it.

src/utils/metadata.ts
```diff
@@ -99,6 +99,22 @@ function getAlgorithmFileExtension(fileUrl: string): string {
   return splitedFileUrl[splitedFileUrl.length - 1]
 }
 
+function sanitizeUrl(url: string): string {
+  return url.replace('javascript:', '')
```
oceanByte, line 103: question(blocking): Does the xss attack still work when the `'javascript:'` is written several times? 
```
javascript:javascript:
```

If pattern in `replace()` is a string, only the first occurrence will be replaced.

oceanByte (approved):

PR #73, opened 2022-03-01
Fix: query for algo's compatible datasets

### LINKED ISSUE #67: [BUG] Whitelisted Datset not shown on Algorithm GUI
For the algorithm BIDS Example Application for CtD
https://portal.minimal-gaia-x.eu/asset/did:op:8eD56443192e1160185922384F20199513FaDEC9 the whitelisted data asset associated with it BIDS Data - Heartbeat Detection Task
https://portal.minimal-gaia-x.eu/asset/did:op:16229fE8e0Ebb09224E21E3dDA23D584Ea33f201 is not shown in the GUI. 

The other way around it is shown. This should be fixed to demonstrate the utility of algorithms and how they are used.

Algo: 
image

Data:
image

LucaMilanese90:
Fixes #67 

## Proposed Changes

  - replace "term" with "match_phrase" in `getAlgorithmDatasetsForCompute` base query
  - hide algorithms with different service endpoint than dataset in ctd asset actions

### DIFF
--- a/src/components/organisms/AssetActions/Compute/index.tsx
+++ b/src/components/organisms/AssetActions/Compute/index.tsx
@@ -179,7 +179,8 @@ export default function Compute({
       // verify if the algorithms in the trusted list share the same endpoint of the dataset
       const sameProviderEndpointAlgorithms = queryResults.results.filter(
         (algo) =>
-          algo.service[1].serviceEndpoint === computeService.serviceEndpoint
+          algo.findServiceByType('compute')?.serviceEndpoint ===
+          computeService.serviceEndpoint
       )
       setDdoAlgorithmList(sameProviderEndpointAlgorithms)
 
--- a/src/components/organisms/AssetActions/Compute/index.tsx
+++ b/src/components/organisms/AssetActions/Compute/index.tsx
@@ -169,21 +169,30 @@ export default function Compute({
     ) {
       algorithmSelectionList = []
     } else {
-      const gueryResults = await queryMetadata(
+      const queryResults = await queryMetadata(
         getQuerryString(
           computeService.attributes.main.privacy.publisherTrustedAlgorithms,
           ddo.chainId
         ),
         source.token
       )
-      setDdoAlgorithmList(gueryResults.results)
-      const datasetComputeService = ddo.findServiceByType('compute')
-      algorithmSelectionList = await transformDDOToAssetSelection(
-        undefined,
-        gueryResults.results,
-        [],
-        newCancelToken()
+      // verify if the algorithms in the trusted list share the same endpoint of the dataset
+      const sameProviderEndpointAlgorithms = queryResults.results.filter(
+        (algo) =>
+          algo.service[1].serviceEndpoint === computeService.serviceEndpoint
       )
+      setDdoAlgorithmList(sameProviderEndpointAlgorithms)
+
+      if (sameProviderEndpointAlgorithms.length === 0) {
+        algorithmSelectionList = []
+      } else {
+        algorithmSelectionList = await transformDDOToAssetSelection(
+          undefined,
+          sameProviderEndpointAlgorithms,
+          [],
+          newCancelToken()
+        )
+      }
     }
     return algorithmSelectionList
   }
--- a/src/components/organisms/AssetActions/Compute/index.tsx
+++ b/src/components/organisms/AssetActions/Compute/index.tsx
@@ -311,7 +311,6 @@ export default function Compute({
       }
 
       if (!hasPreviousDatasetOrder && !hasDatatoken) {
-        console.log(price, ddo)
         const tx = await buyDT('1', price, ddo)
         if (!tx) {
           setError('Error buying datatoken. Please try again.')
@@ -321,7 +320,6 @@ export default function Compute({
       }
 
       if (!hasPreviousAlgorithmOrder && !hasAlgoAssetDatatoken) {
-        console.log(algorithmPrice, selectedAlgorithmAsset)
         const tx = await buyDT('1', algorithmPrice, selectedAlgorithmAsset)
         if (!tx) {
           setError('Error buying datatoken. Please try again.')
--- a/src/components/organisms/AssetActions/Compute/index.tsx
+++ b/src/components/organisms/AssetActions/Compute/index.tsx
@@ -179,8 +179,10 @@ export default function Compute({
       // verify if the algorithms in the trusted list share the same endpoint of the dataset
       const sameProviderEndpointAlgorithms = queryResults.results.filter(
         (algo) =>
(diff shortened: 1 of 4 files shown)


moritzkirstein (changes requested):
Viewing the code now, I believe I already integrated this into our implementation of the `getFilterTerm` function.
Please have a look at this, I think you can already specify to get a `match` object instead of `term` using the function

moritzkirstein:
It seems like this unfortunately introduces some other kind of buggy behavior - probably the reason we changed to `term` here:

grafik

`match` matches datasets that do not actually contain the algo did in the `publisherTrustedAlgorithm` field as it seems. Could you have another look at this?

Looking at one of the listed datasets:
grafik

It does not list the algorithm (correct), so this is only related to algorithms displaying wrong datasets it seems.

LucaMilanese90:
@moritzkirstein I'm reading through the elasticsearch documentation and `match_phrase` (instead of `match`) seems to be the query we are looking for. Do you see any downsides to use it?

LucaMilanese90:
@moritzkirstein The problem seems to be in the `transformDDOToAssetSelection` (line 297) in the case dataset and algo provider endpoint are not matching. In the example you mentioned the dataset has the endpoint https://provider.gaiax-gpu.delta-dao.com/ while the algo has https://provider.safefbdc.delta-dao.com/ .
If I artificially make them the same the algorithm shows up in the list.
I'm not sure what the endpoint purpose is though, and if it's going to create problems if we ignore it.

LucaMilanese90:
@moritzkirstein I've looked through all the repositories but I didn't find any changes that may allow to start ctd with assets having different endpoints.
I also noticed that, even though in the dataset you linked there is an algorithm shown between the possible options, it's not possible to start the job (always throws an error "failed to start job").
Maybe the mistake is that the algo is shown in the dataset tab, not the other way around, does it make sense to you?

src/components/organisms/AssetActions/Compute/index.tsx
```diff
@@ -169,21 +169,30 @@ export default function Compute({
...
-      const datasetComputeService = ddo.findServiceByType('compute')
-      algorithmSelectionList = await transformDDOToAssetSelection(
-        undefined,
-        gueryResults.results,
-        [],
-        newCancelToken()
+      // verify if the algorithms in the trusted list share the same endpoint of the dataset
+      const sameProviderEndpointAlgorithms = queryResults.results.filter(
+        (algo) =>
+          algo.service[1].serviceEndpoint === computeService.serviceEndpoint
```
moritzkirstein, line 182: issue(blocking): There is a helper to get the correct service that we should use. Please have a look at `findServiceByType(type)` -> in this case I believe we are looking for the service of type "compute"

e.g.
```ts
(algo) => algo.findServiceByType('compute')?.serviceEndpoint === computeService.serviceEndpoint
```

moritzkirstein (changes requested):

src/components/organisms/AssetActions/Compute/index.tsx
```diff
@@ -169,21 +169,33 @@ export default function Compute({
...
-        [],
-        newCancelToken()
+      // verify if the algorithms in the trusted list share the same endpoint of the dataset
+      const sameProviderEndpointAlgorithms = queryResults.results.filter(
+        (algo) =>
+          (
+            algo.findServiceByType('compute') ||
+            algo.findServiceByType('access')
+          )?.serviceEndpoint === computeService.serviceEndpoint
       )
```
oceanByte, line 186: nitpick(blocking): This is 'clever' code. It uses different JavaScript features to make the code concise. But you pay with bad readability. Extract the filter into its own const instead of doing everything in one statement.

src/components/organisms/AssetActions/Compute/index.tsx
```diff
@@ -169,21 +169,33 @@ export default function Compute({
...
+        algorithmSelectionList = []
+      } else {
+        algorithmSelectionList = await transformDDOToAssetSelection(
+          undefined,
+          sameProviderEndpointAlgorithms,
+          [],
+          newCancelToken()
+        )
+      }
     }
```
oceanByte, line 199: suggestion(blocking): This looks like there is a cleaner way to set this value. Think of something like this (there might be even better ways using `concat`): 

```javascript
if (sameProviderEndpointAlgorithms.length > 0) {
      algorithmSelectionList = await transformDDOToAssetSelection(
          undefined,
          sameProviderEndpointAlgorithms,
          [],
          newCancelToken()
        )
}
```

src/utils/aquarius.ts
```diff
@@ -35,7 +35,7 @@ export const MAXIMUM_NUMBER_OF_PAGES_WITH_RESULTS = 476
 export function getFilterTerm(
   filterField: string,
   value: string | number | boolean | number[] | string[],
-  key: 'terms' | 'term' | 'match' = 'term'
+  key: 'terms' | 'term' | 'match' | 'match_phrase' = 'term'
 ): FilterTerm {
   const isArray = Array.isArray(value)
   const useKey = key === 'term' ? (isArray ? 'terms' : 'term') : key
```
oceanByte, line 41: nitpick(non-blocking): This looks like 'clever' code. There might be an easier way to express this in more than one line.

oceanByte (changes requested):

src/utils/aquarius.ts
LucaMilanese90: You mean by splitting the options from the default value assignation?

src/components/organisms/AssetActions/Compute/index.tsx
LucaMilanese90: Thought it was fine since this specific shorthand is used multiple times over the rest of the file, I'll change it to something more readable

src/components/organisms/AssetActions/Compute/index.tsx
LucaMilanese90: I ended up refactoring the whole function, it looks better to me now

src/utils/aquarius.ts
moritzkirstein: I thought about this, too: A dedicated type might be useful now.
```ts
type TFilterType = 'terms' | 'term' | 'match' | 'match_phrase'
type TFilterValue = string | number | boolean | number[] | string[]

export function getFilterTerm(
  field: string,
  value: TFilterValue,
  type: 'terms' | 'term' | 'match' = 'term'
) : FilterTerm {}
```

moritzkirstein:
Nice!! Preview looks really good so far 👏

moritzkirstein (approved):

moritzkirstein:
Please review again / resolve @oceanByte

oceanByte (approved):

PR #103, opened 2022-03-25
Remove video player

moritzkirstein:
Fixes #

## Proposed Changes

  - see #64


oceanByte (approved):

LucaMilanese90:
I see the close button is super small when the image is zoomed in, maybe we can remove it completely since the cursor shows the "zoom out" icon to go back

PR #105, opened 2022-03-28
fix: search bar width for desktop view

LucaMilanese90:
## Proposed Changes

  - fix search bar desktop view


moritzkirstein:
There is a small window where the searchbar is positioned awkwardly before being set to full width:

grafik

moritzkirstein (approved):

PR #106, opened 2022-03-28
Fix: modal close button size

### LINKED ISSUE #104: Fix display issue of modal close button
I see the close button is super small when the image is zoomed in on big screens, maybe we can keep it only for mobile views since for the desktops we already have the cursor showing the "zoom out" icon to go back

_Originally posted by @LucaMilanese90 in https://github.com/deltaDAO/mvg-portal/pull/103#issuecomment-1078980937_

LucaMilanese90:
Fixes #104 

## Proposed Changes

  - image doesn't overflow anymore from the modal window
  - adjust close button size to be visible on all screen sizes

### DIFF
--- a/src/components/molecules/InteractiveModalImage.module.css
+++ b/src/components/molecules/InteractiveModalImage.module.css
@@ -34,12 +34,14 @@
 
 #close:before,
 #close:after {
+  --close-width: 0.125rem;
+
   position: absolute;
   top: 15%;
-  left: calc(50% - 0.0625rem);
-  width: 0.125rem;
+  left: calc(50% - (var(--close-width) / 2));
+  width: var(--close-width);
   height: 70%;
-  border-radius: 0.125rem;
+  border-radius: var(--close-width);
   transform: rotate(45deg);
   background: currentcolor;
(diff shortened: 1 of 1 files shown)


src/components/molecules/InteractiveModalImage.module.css
```diff
@@ -31,10 +36,10 @@
 #close:after {
   position: absolute;
   top: 15%;
-  left: calc(50% - 0.0625em);
-  width: 0.125em;
+  left: calc(50% - 0.0625rem);
+  width: 0.125rem;
   height: 70%;
-  border-radius: 0.125em;
+  border-radius: 0.125rem;
```
moritzkirstein, line 42: suggestion(blocking): I would prefer having this number in a css variable. Can be scoped to only this close btn if we don't use it anywhere else, e.g.
```css
#close:after {
  --close-width: 0.125rem;
  width: var(--close-width);
  left: calc(var(--close-width) / 2);
}
```

moritzkirstein (changes requested):

moritzkirstein (approved):

Issue #113, opened 2022-04-07
[Feature] Onboarding Steps


moritzkirstein:
## Motivation / Problem

Onboarding Step components should be implemented according to #107

## Solution

Implement one component per step that can be displayed by the onboarding section component, see #112 
- [ ] Step 1: MetaMask
- [ ] Step 2: Connect account
- [ ] Step 3: Connect to network
- [ ] Step 4: Add testnet tokens
- [ ] Step 5: Claim testnet tokens
- [ ] Step 6: Success

## Additional context

- Status checks for each step should be done dynamically by the step components (e.g. check if user is connected to gx testnet)
- Step component should support optional images (screenshots) to be displayed according to figma design

PR #115, opened 2022-04-12
Feat: onboarding section

### LINKED ISSUE #112: [Feature] Onboarding Section
## Motivation / Problem

We want to implement a new onboarding feature according to #107 

## Solution

Add a new section to the landing page which will hold the n-step onboarding as defined in #107 
Section should be composed of
- [ ] Intro header (text)
- [ ] progress bar
- [ ] current step component (see #113)
- [ ] step "pagination"

## Additional Context
- The section should be a new component 
- Design can be found in #110 
- The current step # should be stored in and retrieved from local storage (maybe make use of existing UserPreferences provider)
- Enable support for n steps (easy addition / deletion of steps in the future)

LucaMilanese90:
Fixes #112 #113

## Proposed Changes

  - implement onboarding section
  - enable support for undefined amount of steps
  - automatic check if onboarding step has been completed
  - step layout with optional picture


content/pages/index/onboarding/steps/ready.json
```diff
@@ -0,0 +1,5 @@
+{
+  "title": "Good job! You are ready to explore the portal, scroll down to see more.",
+  "suggestion": "One last thing: we are never going to ask for your private key,\n\n remember to never share it with anyone!",
```
oceanByte, line 3: nitpick(blocking): do not couple the variable name too much to the content. It's a body text after all.

src/components/atoms/Button.module.css
```diff
@@ -66,6 +67,23 @@
...
 
+.outline {
+  background: linear-gradient(
+    to right,
+    var(--color-primary) 50%,
+    var(--brand-white) 50%
+  );
+  background-size: 202% 100%;
+  background-position: bottom right;
+  border: 2px solid var(--color-primary);
```
oceanByte, line 78: question(non-blocking): Is there some global value for border thickness? It would be great to be consistent here and allow for an easy update.

src/components/atoms/Button.module.css
```diff
@@ -66,6 +67,23 @@
   background: var(--brand-white);
 }
 
+.outline {
+  background: linear-gradient(
+    to right,
+    var(--color-primary) 50%,
+    var(--brand-white) 50%
+  );
+  background-size: 202% 100%;
```
oceanByte, line 76: question(non-blocking): I get suspicious when I read magic numbers like 202. Can we try to work with multiples of a base value?

src/components/organisms/Onboarding/StepActions.tsx
```diff
@@ -0,0 +1,42 @@
...
+  buttonAction,
+  successMessage,
+  loading,
+  completed,
+  loadingMessage
+}: IStepAction): ReactElement {
+  return (
+    <div className={styles.container}>
+      {loading ? (
+        <Loader message={loadingMessage || 'Loading...'} />
```
oceanByte, line 28: nitpick(non-blocking): We could also use a dedicated variable here that contains the `'Loading...'` text. It's cleaner.

src/components/organisms/Onboarding/StepActions.tsx
```diff
@@ -0,0 +1,42 @@
...
+      ) : completed ? (
+        <Alert
+          text={successMessage}
+          state="success"
```
oceanByte, line 32: nitpick(non-blocking): See the previous comment. Think about using an enum here.

src/components/organisms/Onboarding/StepBody.module.css
```diff
@@ -0,0 +1,51 @@
...
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
+  align-items: center;
+  max-width: 20rem;
+  min-height: 20rem;
+  height: 100%;
+}
+
+@media screen and (min-width: 60rem) {
```
oceanByte, line 42: question(blocking): Does the order follow any logic? If not, I'd just go for an alphabetic order here.

src/components/organisms/Onboarding/StepHeader.tsx
```diff
@@ -0,0 +1,17 @@
...
+  title,
+  subtitle
+}: {
+  title: string
+  subtitle: string
+}): ReactElement {
+  return (
+    <div className={styles.header}>
+      <h3 className={styles.title}>{title}</h3>
+      <h5 className={styles.subtitle}>{subtitle}</h5>
```
oceanByte, line 14: question(blocking): I assume `h4` is used for something in-between title and subtitle? If possible we should change this accordingly though. It is not reasonable that the subtitle differs by 2 here.

src/components/pages/Home/Onboarding/Header.module.css
```diff
@@ -0,0 +1,23 @@
...
+  padding-left: 0;
+}
+
+.content {
+  padding: var(--spacer);
+}
+
+.subtitle {
+  display: inline-block;
+  padding: calc(var(--spacer) / 8) calc(var(--spacer) / 4) 0;
```
oceanByte, line 15: nitpick(blocking): this looks like I need to think a lot about padding and why these numbers were chosen. Add a comment explaining why exactly these values are needed or use variable names.

src/components/pages/Home/Onboarding/Header.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export default function Header(): ReactElement {
+  const data: OnboardingHeaderData = useStaticQuery(query)
+  const { title, subtitle, body } = data.file.childOnboardingJson
+
+  return (
+    <Container className={styles.container}>
+      <div className={styles.content}>
+        <h5 className={styles.subtitle}>{subtitle}</h5>
+        <h2 className={styles.title}>{title}</h2>
```
oceanByte, line 37: nitpick(blocking): Previously title was treated as `h3`. Is there a way to unify this? I am asking since the subtitle is still `h5` here, but if you'd followed the previous logic it should be `h4`.

src/components/pages/Home/Onboarding/Main.tsx
```diff
@@ -0,0 +1,54 @@
...
+    },
+    enter: { opacity: 1, transform: 'translate3d(0%,-50%,0)' },
+    leave: {
+      opacity: 0,
+      transform: `translate3d(${
+        navigationDirection === 'prev' ? '50%' : '-50%'
+      },-50%,0)`
+    },
+    config: { mass: 1, tension: 140, friction: 18 }
+  })
```
oceanByte, line 36: issue(blocking): readability could be improved here. Maybe it helps if we use values like "left", "right" and so on instead of the pure numbers.

src/components/pages/Home/Onboarding/Stepper.tsx
```diff
@@ -0,0 +1,69 @@
...
+          networkId,
+          balance
+        })
+      )
+      return
+    }
+
+    currentStep > newStep
+      ? setNavigationDirection(NavigationDirections.PREV)
+      : setNavigationDirection(NavigationDirections.NEXT)
```
oceanByte, line 42: nitpick(blocking): This piece of logic should be not handled by the clicker but by a dedicated function. Abstraction. Try to keep the logic of the click handler to a high level. It sets the navigation direction and sets the current step. Currently, you are also calculating which step to set. This should be none of your business.

src/components/pages/Home/Onboarding/Steps/ClaimTokens.tsx
```diff
@@ -0,0 +1,157 @@
...
+      // the api call is going to fail due to a CORS error but the tokens are
+      // sent anyway so we set the new token state in the catch
+      setTokenState({
+        ...tokenState,
+        [token]: {
+          ...tokenState[token],
+          completed: true,
+          touched: true,
+          loading: false
+        }
```
oceanByte, line 129: issue(blocking): How do you deal with the case if someone already claimed the token for that day? Is this ignored for v1?

src/utils/onboarding.ts
```diff
@@ -0,0 +1,35 @@
...
+    return 'Looks like you are not connected to the Gaia-X testnet, please go back to the "Network" step.'
+  }
+  if (balance !== null) {
+    if (Number(balance?.eth) === 0) {
+      return "Looks like you don't have any GX test tokens, please click on the claim button to get some."
+    }
+    if (Number(balance?.ocean) === 0) {
+      return "Looks like you don't have any OCEAN test tokens, please click on the claim button to get some."
+    }
+  }
```
oceanByte, line 30: nitpick(non-blocking): Duplication. Consider using variables for the words that change, but it's not a lot of text so do not overengineer here. It's optional for now.

src/components/pages/Home/Onboarding/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+  NEXT = 'next'
+}
+
+export default function OnboardingSection(): ReactElement {
+  const { accountId, balance, networkId, web3Provider } = useWeb3()
+  const { onboardingStep, setOnboardingStep } = useUserPreferences()
+  const [onboardingCompleted, setOnboardingCompleted] = useState(false)
+  const [navigationDirection, setNavigationDirection] =
+    useState<NavigationDirections>()
+  const stepLabels = steps?.map((step) => step.shortLabel)
```
oceanByte, line 53: question(non-blocking): when could `steps` be empty or undefined?

oceanByte (changes requested):

src/components/pages/Home/Onboarding/Stepper.tsx
```diff
@@ -0,0 +1,69 @@
...
+            key={i}
+            className={cx({
+              step: true,
+              active: i <= currentStep
+            })}
+          >
+            <Button
+              style="text"
+              className={styles.title}
+              onClick={() => handleClick(i)}
```
moritzkirstein, line 60: thought/nitpick(non-blocking): usability / accessibility of this component would be imrpoved if the "step circles" could also be clicked imo

content/pages/index/onboarding/steps/claimTokens.json
```diff
@@ -0,0 +1,12 @@
...
+  "title": "Get some test tokens",
+  "subtitle": "To publish or consume the assets in the portal you'll need some GX and OCEAN test tokens to spend",
+  "body": "Click on the buttons below to claim your test tokens.",
+  "image": "../../../../../src/images/onboarding/claim_tokens.png",
+  "gxButtonLabel": "Claim 1 GX",
+  "existingGxBalance": "You already have some GX test tokens in your wallet 🎉",
+  "gxSuccess": "Successfully claimed 1 GX",
+  "oceanButtonLabel": "Claim 100 OCEAN",
+  "existingOceanBalance": "You already have some OCEAN test tokens in your wallet 🎉",
+  "oceanSuccess": "Successfully claimed 100 OCEAN"
```
moritzkirstein, line 11: thought(blocking): This could be formatted more elegantly with an array:
```json
"buttons": [
  {
    "name": "gxButton",
    "label": "Claim 1 GX",
    "balance": "You already have some GX test tokens in your wallet 🎉",
    "success": "Successfully claimed 1 GX"
  },
  {
    ...
  }
]
```

This also enables a "better" query in the component (see below)

src/components/pages/Home/Onboarding/Steps/ClaimTokens.tsx
```diff
@@ -0,0 +1,157 @@
...
+              src
+            }
+          }
+        }
+        gxButtonLabel
+        existingGxBalance
+        gxSuccess
+        oceanButtonLabel
+        existingOceanBalance
+        oceanSuccess
```
moritzkirstein, line 34: issue(blocking): A very long query for this component. It seems like this can be simplified, e.g:
```jsx
const query = graphql`
  query ClaimTokensQuery {
    file(
      relativePath: { eq: "pages/index/onboarding/steps/claimTokens.json" }
    ) {
      childStepsJson {
        title
        subtitle
        //...
        buttons {
          name
          label
          balance
          success
       // ...
```

Enables adding additional buttons easier as a bonus

src/components/organisms/Onboarding/StepActions.tsx
```diff
@@ -0,0 +1,42 @@
...
+interface IStepAction {
+  buttonLabel: string
+  buttonAction: (e: FormEvent) => void
+  successMessage?: string
+  loading: boolean
+  completed: boolean
+  loadingMessage?: string
+}
+
+export default function StepActions({
```
moritzkirstein, line 17: nitpick(non-blocking): Component always renders 1 action. It should be named `StepAction` isntead

src/components/organisms/Onboarding/StepBody.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { ReactNode } from 'react-markdown'
+
+export default function StepBody({
+  body,
+  image,
+  children
+}: {
+  body: string
+  image?: string
+  children?: ReactNode
```
moritzkirstein, line 13: thought(non-blocking): I think using children here does potentially introduce some confusion. It communicates that any elements can be passed, even though it is only used to display the steps actions.

I feel like the better approach would be to pass the actions to the component, like:
```jsx
export default function StepBody({
  body,
  image,
  actions
}: {
  body: string
  image?: string
  actions?: IStepAction[]
})
```

and then display `StepAction`s for each action passed

src/components/atoms/Button.module.css
LucaMilanese90: 200% was working fine on Chrome but Firefox for some reason was showing a little white space at the end, not sure how to fix the problem differently. I could put a bigger round number (like 300%) but I don't think it would make it clearer and that would also affect the transition speed, would this option be better in your opinion?

src/components/pages/Home/Onboarding/Steps/ClaimTokens.tsx
LucaMilanese90: Right now we are not showing the button to claim tokens at all if you already have some tokens in your wallet, this is also the behaviour of the faucet as far as I know. 
Also being this tool aimed for the first onboarding it's unlikely the case of a user who got some tokens from the faucet, spent them all and after that came to the portal to get more through our guided process.
We can probably handle this easily when we'll have our own implementation of the faucet but I personally think it's out of scope for this v1.

src/components/pages/Home/Onboarding/index.tsx
LucaMilanese90: It's a leftover from the previous implementation when it was passed as a prop, my bad.

src/components/atoms/Button.module.css
oceanByte: Thanks for the clarification. Adding a comment explaining that this is a firefox specific fix should be sufficient then.

src/components/pages/Home/Onboarding/Steps/ClaimTokens.tsx
oceanByte: Sounds good enough for v1. Please create an issue so that we can tackle this in our next release.

src/components/organisms/PrivacyPreferenceCenter.module.css
```diff
@@ -105,6 +105,7 @@
   left: -25rem;
   opacity: 0;
   box-shadow: none;
+  pointer-events: none;
```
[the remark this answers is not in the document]
LucaMilanese90: I needed to change this because it was very hard to click on the previous step button on the laptop screen, but you are right, should I open a separate PR then?

src/components/pages/Home/Onboarding/Stepper.tsx
LucaMilanese90: We thought about it but the connecting lines between the circles are part of the same component and they would be clickable as well, not sure I have a solution atm to make only the circles clickable

src/components/organisms/Onboarding/StepActions.tsx
LucaMilanese90: We added the default message in the StepAction component instead of the Loader itself because in the other way we would lose the ability to show the loader animation by itself, without any text. 
We could add another prop to the Loader component to handle the case, like an optional boolean to decide to not show the default message, but that would require a compatibility check over the whole portal to ensure we are not modifying any existing behaviour. It looked more practical to only add it to this component given the situation, but we can do it if you think it's going to be better

src/components/atoms/Button.module.css
LucaMilanese90: There is only a variable for the border radius, not for the border size. Probably because all the other button styles are color filled and so they don't have a border. It would be a variable used mainly in this component and maybe in the navbar, but we can create it anyway, no problem

src/components/organisms/Onboarding/StepBody.module.css
LucaMilanese90: They are supposed to be ordered by the tag depth, but there is a mistake since `card` should come before `actions`. The order should be `content`, then `image` and `cardContainer` at the same level, then `card` and at the end `actions`.
Maybe it does not make much sense though and we should still switch to the alphabetical order

src/components/atoms/Button.module.css
oceanByte: Ideally, these values would follow global values. A designer taught me that good design also tries to minimize variation and only adds it if there is no alternative or better said if it improves the overall design.

It's a detail, but I'd go for the variable here and replace it for the navbar and this case.

src/components/pages/Home/Onboarding/Header.module.css
LucaMilanese90: We took the style from the component inside the hero section to keep it consistent, so I'm not sure about the reason of these specific padding combination. We could probably extract them into a separate component to keep the style consistent over the page

src/components/organisms/Onboarding/StepBody.module.css
oceanByte: I am fine with both, just make sure that it follows a given logic. 
Hint: Alphabetical order is easier to understand for someone who wants to enhance this file and does not know anything about it's usage yet.

LucaMilanese90: We are not enforcing any convention as far as I know, but looking at the preexisting code it looks like there is an attempt to follow the tag depth convention, that's the reason I went with it. I agree that it's difficult to maintain over time and in a new repo I would probably go alphabetically, but I'm not sure if we should use it here since it would not be consistent with the rest of the codebase

oceanByte: Agreed. And there are reasons for both attempts. Just fix the mistake and we are good to go here.

src/components/pages/Home/Onboarding/Steps/ClaimTokens.tsx
LucaMilanese90: Done: #120

src/components/pages/Home/Onboarding/Stepper.tsx
moritzkirstein: Can be an improvement in a later version if needed +1

src/components/organisms/PrivacyPreferenceCenter.module.css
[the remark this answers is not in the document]
moritzkirstein: Its fine for now from my pov - just keep it in mind for further PRs

src/components/organisms/Onboarding/StepActions.tsx
moritzkirstein: Did not think of the display w/o message. Good catch! I do think however, that this is a functionality that is directly related to and should be provided by the `Loader` component itself. Have a boolean `showMessage` that is set to true by default for example.

src/components/pages/Home/Onboarding/Header.tsx
LucaMilanese90: My impression is that in the whole repo we are using the <h> tags more for the size than for the meaning of the tags. We also have some css global rules to modify the size of the <h> tags based on the screen size, if we use an <h3> instead of an <h2> then we need to "force it" to a different font size in css and put the media query inside the specific component. We can do it, it's just probably impractical to keep updated in the future

src/components/pages/Home/Onboarding/Main.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { animated, useSpringRef, useTransition } from 'react-spring'
+import styles from './Main.module.css'
+
+const translateMovements = {
+  fromTranslateLeft: 'translate3d(-100%,-50%,0)',
+  fromTranslateRight: 'translate3d(100%,-50%,0)',
+  leaveTranslateLeft: 'translate3d(50%,-50%,0)',
+  leaveTranslateRight: 'translate3d(-50%,-50%,0)',
+  startPosition: 'translate3d(0%,-50%,0)'
+}
```
oceanByte, line 11: praise: much easier to read!

src/components/atoms/Loader.tsx
```diff
@@ -2,14 +2,18 @@ import React, { ReactElement } from 'react'
 import styles from './Loader.module.css'
 
 export default function Loader({
-  message
+  message,
+  spinnerOnly
 }: {
   message?: string
+  spinnerOnly?: boolean
```
moritzkirstein, line 9: question(blocking): Are we sure this didn't break anything as you mentioned before?
The safe change would have been to set this to true by default, so we do not have to adjust existing Loaders where this would be the expected behaviour of the component, right?

moritzkirstein: edit: I see you went through the codebase adjusting this where necessary. Still: It is difficult to tell if there are possible cases that were missed and the workload was likely unecessary. At least for future adjustments like this, we need to make sure to try to stay away from introducing any new behaviour to components that require adjustments in the whole codebase.

moritzkirstein (changes requested):
One thought / question left to resolve. Otherwise great work! Once this has been clarified we can move ahead with the merge.

src/components/atoms/Loader.tsx
LucaMilanese90: I went for this option because imho it looks more natural to pass a flag to modify some behaviour rather than pass a  false value. 
I do agree on the reasoning behind your comment, so I've changed them back, but I still think the false default is easier to read

moritzkirstein (approved):

src/components/atoms/Loader.tsx
LucaMilanese90: Updated the component to have an optional `useDefaultMessage` flag (false by default) instead of `spinnerOnly`

moritzkirstein (approved):

PR #119, opened 2022-04-25
Fix: buy correct datatoken for compute job

### LINKED ISSUE #126: [BUG] Failed to start compute job
## Summary
In the asset detail page, selecting an algorithm and pressing on the "buy compute job" button sometimes leads to an error.

image

## Steps to Reproduce
1. select a dataset from the compute section
2. select an algorithm
3. quickly change your selection with another algorithm and press the "buy" button
  
## Proposed solution
Disable algorithm selection and show a loader until price fetch completion, this should avoid the problem.

LucaMilanese90:
Fixes #126

## Proposed Changes

  - disable algorithm asset list while fetching the selected algorithm price to avoid buying the wrong data token

### DIFF
--- a/src/components/organisms/AssetActions/Compute/PriceOutput.tsx
+++ b/src/components/organisms/AssetActions/Compute/PriceOutput.tsx
@@ -73,7 +73,7 @@ export default function PriceOutput({
     <div className={styles.priceComponent}>
       {loading ? (
         <div className={styles.loader}>
-          <Loader message="Fetching asset total price" />
+          <Loader message="Fetching assets total price" />
         </div>
       ) : (
         <>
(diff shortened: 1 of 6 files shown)


src/components/organisms/AssetActions/Compute/PriceOutput.tsx
```diff
@@ -62,35 +64,45 @@ export default function PriceOutput({
   hasPreviousOrderSelectedComputeAsset,
   hasDatatokenSelectedComputeAsset,
   algorithmPrice,
-  selectedComputeAssetTimeout
+  selectedComputeAssetTimeout,
+  loading
```
oceanByte, line 68: nitpick(blocking): We should follow the convention for booleans here and use `isLoading` instead.

oceanByte (approved):

moritzkirstein (approved):

PR #144, opened 2022-06-30
Enhancement: menu dropdown

### LINKED ISSUE #118: [Enhancement] Improve Header and Footer with Feature Links
## Motivation / Problem
It is currently difficult for users to find all the resources of the Minimal Viable Gaia-X as the links are not being presented on the landing page. This should be as easy as possible for users and first-time visitors.

## Solution 
Include links to the following resources in the top and footer navigation:

- Catalogue https://catalogue.minimal-gaia-x.eu/
- Exchange Logging https://exchangelog.minimal-gaia-x.eu/
- Stats Page https://stats.minimal-gaia-x.eu/
- Onboarding https://onboarding-portal.lab.gaia-x.eu/
- Additonal on footer: Link to the Gaia-X Page https://gaia-x.eu/

Challenge: Link texts need to very brief, so i propose:

Catalogue, Logging, Stats, Identity for the top navigation. I am aware that this might create issues with the mobile navigation thus moving the links in the upper section of the page might be an alternative to the top navigation.

Top navigation
image

Alternative
image

Footer (does not replace the top nav, we need the links in the top nav)
image

## Alternatives
Create a section like the features assets section that leads to the additional pages and teasers them briefly and describes their functionality.

## Additional context
User centricity needs to be in focus. As we lack a structured user testing panel we need to change the perspective to the one of a newcomer frequently.

LucaMilanese90:
Fixes #118 

## Proposed Changes

  - Add additional navbar menu item with support for sub-items


moritzkirstein (commented):
Looks good to me. We may have to update this for mobile screens a bit, will check more later. It looks like the dropdown (especially 2nd layer) sometimes can get out of bounds a bit.

LucaMilanese90:
@moritzkirstein Yes, on very small screens it can be a problem, I've changed it to an automatic placement, it looks a little worse on desktop but at least it's always visible on mobile

src/components/atoms/MenuDropdown.tsx
```diff
@@ -0,0 +1,83 @@
...
+                      {item.subItems.map((subItem, i) => (
+                        <li key={`${subItem.name}-${i}`}>
+                          <LinkOpener
+                            uri={subItem.link}
+                            className={styles.link}
+                          >
+                            {subItem.name}
+                          </LinkOpener>
+                        </li>
+                      ))}
```
moritzkirstein, line 60: suggestion(blocking): I feel like this could probably be refactored into its own component. Something like:
```jsx
<ul>
  <li key={`${item.name}-${i}`}>
    {item.subItems ?
      <Tooltip ... content={<ItemComponent subitems={item.subitems}} /> :
      <LinkOpener
        uri={item.link}
        className={styles.link}
      >
        {item.name}
      </LinkOpener>
    }
  </li>
</ul>

```

Did you consider this, or could have a look what is possible here to reduce the code duplication?

moritzkirstein (changes requested):

moritzkirstein (approved):
If all works as expected and you agree with the proposed changes you can go ahead with the merge @LucaMilanese90

LucaMilanese90:
Yep, it works correctly, I like it😄

PR #146, opened 2022-07-05
enhancement: update imprint content

LucaMilanese90:


moritzkirstein (approved):
question(non-blocking): Do we want to order members of the board alphabetically? @fr-3deric @kaimeinke @oceanByte

LucaMilanese90:
Frederic and Kai said the order is not important, I think we can merge the PR

PR #150, opened 2022-07-07
feat: add new featured assets

Abrom8:
## Proposed Changes

  - change featured assets
  - extend to show 9 instead of 6

### DIFF
--- a/src/components/pages/Home.tsx
+++ b/src/components/pages/Home.tsx
@@ -244,16 +244,12 @@ export default function HomePage(): ReactElement {
         query: generateBaseQuery({ ...baseParams, ...queryParams })
       })
     }
-    if (featuredSections.length > 0) {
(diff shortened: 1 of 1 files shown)


src/components/pages/Home.tsx
```diff
@@ -209,20 +260,24 @@ export default function HomePage(): ReactElement {
...
+                title={section.title}
+                query={section.query}
+              />
+            ))}
+          <Button
+            className={styles.allAssetsButton}
+            style="text"
+            to="/search?sort=created&sortOrder=desc"
+          >
+            All data sets and algorithms →
```
oceanByte, line 279: thought(non-blocking): the Unicode Rightwards Arrow character might not be supported by many fonts. So I would rather use a CSS or SVG icon here.

oceanByte (approved):

PR #154, opened 2022-07-07
feat: add verified service provider name to asset details page

LucaMilanese90:
## Proposed Changes

  - show provider name from a verified SD (if available) in asset details
  - detach metadata loading from service SD verification in asset details page

### DIFF
--- a/src/utils/metadata.ts
+++ b/src/utils/metadata.ts
@@ -18,6 +18,7 @@ import {
   EditableMetadataLinks
 } from '@oceanprotocol/lib'
 import { complianceUri } from '../../app.config'
+import { isSanitizedUrl } from '../components/molecules/FormFields/URLInput/Input'
 
 export function transformTags(value: string): string[] {
   const originalTags = value?.split(',')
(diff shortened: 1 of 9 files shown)


src/utils/metadata.ts
```diff
@@ -216,6 +216,14 @@ export function updateServiceSelfDescription(
   return ddo
 }
 
+export function getPublisherFromServiceSD(serviceSD: any): string {
+  if (!serviceSD) return
+
+  return serviceSD?.selfDescriptionCredential?.credentialSubject?.[
+    'gx-service-offering:name'
+  ]?.['@value']
```
moritzkirstein, line 224: issue(blocking): we will need to use another field here I think. Will check with some of our SDs to confirm.

moritzkirstein: We need to make use of the following:
```json
 "credentialSubject": {
          "id": "https://delta-dao.com/.well-known/servicePortal.json",
          "gx-service-offering:providedBy": {
              "@type": "xsd:string",
              "@value": "https://delta-dao.com/.well-known/participant.json"
          }
  //...
```

Load participant SD:
e.g. https://delta-dao.com/.well-known/participant.json
```json
{
  "selfDescriptionCredential": {
    "credentialSubject": {
      "gx-participant:name": {
        "@value": "deltaDAO AG",
        "@type": "xsd:string"
      }
  }
}
```
display this name as publisher

LucaMilanese90: I updated the current behavior to separate the metadata loading from the SD verification. 
As for the service provider name, I think we decided with @kaimeinke during the hackathon to use `gx-service-offering:name`, but we can of course update this if it's not correct. We'll have to mirror this to the catalogue as well since we are using the same field over there. 

> Load participant SD:
e.g. https://delta-dao.com/.well-known/participant.json
if service-offering was verified, we can expect a valid participant SD. No need to verify this again: 

I'm confused on this point: I thought we agreed on performing a verification of every page refresh to make the whole thing more transparent. Also this would still need to happen for the URL versions or the SD since the content could change over time.

moritzkirstein: > I'm confused on this point: I thought we agreed on performing a verification of every page refresh to make the whole thing more transparent. Also this would still need to happen for the URL versions or the SD since the content could change over time.

Sorry, maybe was not clear enough on this. I meant once the compliance check was successfull, we can trust that the URL to the participant SD is conforming. No need to do an **additional** check at that point.

moritzkirstein (approved):

PR #156, opened 2022-07-12
fix: correctly set provider name from raw SD

LucaMilanese90:
## Proposed Changes

  - correctly parse the raw service SD to extract the verified provider name

### DIFF
--- a/src/utils/metadata.ts
+++ b/src/utils/metadata.ts
@@ -221,10 +221,10 @@ export async function getPublisherFromServiceSD(
   serviceSD: any
 ): Promise<string> {
   if (!serviceSD) return
-  const parsedServiceSD =
-    typeof serviceSD === 'string' ? JSON.parse(serviceSD) : serviceSD
 
   try {
+    const parsedServiceSD =
+      typeof serviceSD === 'string' ? JSON.parse(serviceSD) : serviceSD
     const providedByUrl =
       parsedServiceSD?.selfDescriptionCredential?.credentialSubject?.[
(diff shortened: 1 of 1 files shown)


src/utils/metadata.ts
```diff
@@ -221,11 +221,15 @@ export async function getPublisherFromServiceSD(
   serviceSD: any
 ): Promise<string> {
   if (!serviceSD) return
+  const parsedServiceSD =
+    typeof serviceSD === 'string' ? JSON.parse(serviceSD) : serviceSD
```
oceanByte, line 225: issue(blocking): Move this line into the try block to handle JSON.parse errors

oceanByte (changes requested):

oceanByte (approved):

src/utils/metadata.ts
```diff
@@ -221,11 +221,15 @@ export async function getPublisherFromServiceSD(
...
+
   try {
+    const parsedServiceSD =
+      typeof serviceSD === 'string' ? JSON.parse(serviceSD) : serviceSD
     const providedByUrl =
-      serviceSD?.selfDescriptionCredential?.credentialSubject?.[
+      parsedServiceSD?.selfDescriptionCredential?.credentialSubject?.[
         'gx-service-offering:providedBy'
       ]?.['@value']
+
```
moritzkirstein, line 232: suggestion: see https://github.com/deltaDAO/mvg-catalogue/pull/29#discussion_r918921369

Issue #128, opened 2022-05-17
[Enhancement] Refactor CSS color variables


moritzkirstein:
## Motivation / Problem
Issue raised here: https://github.com/deltaDAO/portal-euprogigant/pull/1/files#r874571899

## Solution 
Refactor CSS color variables to reflect the purpose of the color, rather than the "actual" color as this leads to confusion, especially in branded forks.

PR #167, opened 2022-07-26
Enhancement: Improve Loading Speed

### LINKED ISSUE #165: [Enhancement]: Improve loading speed of asset teaser /& details
## Motivation / Problem
The current loading times for asset sections of the portal are too long. We want to improve usability by decreasing the time the user spends on an empty page.

## Solution

**Verify current behaviour:**
A loader is displayed until all asset information is available including calls to aquarius, subgraph etc.

**Expected or prferred behaviour would be to** 
1. load DDOs from aqua
2. prepare DDOs to be in correct format for teasers
3. display teasers immediately with title, desc etc.
4. display spinner / placeholder / nothing for prices account names etc
5. load prices via subgraph
6. load accounts via 3box
7. display info whenever ready

Going from initial network data it seems we could reduce initial loading times for teaser lists & details to about .5s max instead of multiple seconds.

moritzkirstein:
Fixes #165 

## Proposed Changes

  - add gradient style to loader component
  - allow AssetTeaser to display unpriced assets
  - move price loading out of `loading` scope for asset provider

Could think about introducing smth like https://github.com/danilowoz/react-content-loader to improve the look of the loading status.

### DIFF
--- a/src/components/atoms/Loader.module.css
+++ b/src/components/atoms/Loader.module.css
@@ -49,7 +49,7 @@
   animation-iteration-count: infinite;
   animation-name: placeHolderShimmer;
   animation-timing-function: linear;
-  background: darkgray;
+  background: var(--brand-grey-dimmed);
   background: linear-gradient(
     to right,
     var(--brand-grey-dimmed) 10%,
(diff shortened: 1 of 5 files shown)


LucaMilanese90 (approved):
Much faster than before, we can refactor a few CSS variables when we'll work on https://github.com/deltaDAO/mvg-portal/issues/128

oceanByte (approved):

PR #173, opened 2022-08-03
feat: add new searchBar

### LINKED ISSUE #166: Feature: Add search bar to search & landing page
### **Motivation**

Based on feedback, the search bar is often used by our portal users, but right now too hard to find.

### **Solution**:

Add search bar to search & landing page.

Image

Similar implementation as on the federated catalogue page:

Image

Abrom8:
Fixes #166 

## Proposed Changes

  - add searchBar to home and search page


moritzkirstein (changes requested):
I feel like the SearchBar is a little to "unpresesnt" at the home page. Moving it to the right on the search page worked for me. The component may need to support different sizes so it gets more space on landing.

issue(blocking): We need to have the search "button" added back to the header for all other pages though. Search functionality is a major feature users are utilizing and it needs to be available easily from everywhere.

src/components/templates/Page.tsx
```diff
@@ -21,9 +22,15 @@ export default function Page({
   headerCenter
 }: PageProps): ReactElement {
   const isHome = uri === '/'
+  const isSearch = uri === '/search'
 
   const childElements = (
     <>
+      {isSearch && (
+        <>
+          <SearchBar visibleInput placeholder="Search for service offerings" />
```
oceanByte, line 31: nitpick(blocking): Can we get this placeholder value from a central place? We will forget to update this if we write a custom text here.

oceanByte (changes requested):

oceanByte:
image

issue(blocking): Clicking the search button in the navbar results in a screen-wide search bar hovering above everything else. I would either scroll down to the new search bar or find a better way to place the search bar.

Abrom8:
> image
> 
> issue(blocking): Clicking the search button in the navbar results in a screen-wide search bar hovering above everything else. I would either scroll down to the new search bar or find a better way to place the search bar.

That's also the current behavior in our production portal and the hovering is wanted behavior as far as I know. We can change this but I would consider this a new feature/issue.

moritzkirstein:
> > image
> > issue(blocking): Clicking the search button in the navbar results in a screen-wide search bar hovering above everything else. I would either scroll down to the new search bar or find a better way to place the search bar.
> 
> That's also the current behavior in our production portal and the hovering is wanted behavior as far as I know. We can change this but I would consider this a new feature/issue.

Yes, also wanted to add this. This is current behaviour. I think if we are tackling the searchbar now we should take care of it. Got this feedback already a few times that this was bugging people. I think we can fix it in this PR, seems related enough.

oceanByte:
> That's also the current behavior in our production portal and the hovering is wanted behavior as far as I know. We can change this but I would consider this a new feature/issue.

That's correct. But I think we would handle this here as well as we are already working on the search bar. People will click this button if they see the new component and we should ensure a consistent experience.

LucaMilanese90:
Nice, now it looks way better than before! A small thing I noticed from the preview: the first click on the search icon makes the searchbar visible, I would expect that clicking again would hide it, instead it does a weird hide/show animation.

moritzkirstein:
Yes agreed. Way better now that it is not hiding header content.
Would like to have the distance between searchbar and nav a little smaller.
**Current**
grafik

**Example**
grafik

moritzkirstein (changes requested):
Yes agreed. Way better now that it is not hiding header content.
Would like to have the distance between searchbar and nav a little smaller.
**Current**
grafik

**Example**
grafik

src/components/molecules/SearchButton.tsx
```diff
@@ -0,0 +1,34 @@
...
+    }
+  }, [isSearchBarVisible, isHome])
+
+  async function handleButtonClick(e: FormEvent<HTMLButtonElement>) {
+    e.preventDefault()
+    setSearchBarVisible(!isSearchBarVisible)
+    const searchFormSection = document?.getElementById('searchFormSection')
+    if (searchFormSection) {
+      document.getElementById('searchForm').focus()
+      searchFormSection.scrollIntoView()
```
LucaMilanese90, line 23: Thought (non-blocking): to select specific DOM elements in the "React way" you can take advantage of the ref hook https://it.reactjs.org/docs/hooks-reference.html#useref and then pass it to child components with https://reactjs.org/docs/forwarding-refs.html
In this specific case though, you could probably avoid using any ref at all and simply start the scrolling from within the SearchForm component, reacting to the value change of `isSearchBarVisible` inside a useEffect.

moritzkirstein: Agreed with the suggestion. Since we are using the UserPreference provider already, I would suggest adjusting this to use the `useEffect` hook as suggested by @LucaMilanese90 
Meaning you can drop the `id` in `<SearchForm>` and listen to changes of `isSearchBarVisible` in that component instead. The `useRef` still makes sense to perform the scroll itself even from within the component.

moritzkirstein (changes requested):
See comment by @LucaMilanese90 
Should be a good learning experience, further adapting the react way here. I like the new feel & look of the searchbar - nice! 👍

moritzkirstein (commented):
Looks and feels very nice now. I like the new behaviour.
We can go forward with merging to dev from my pov. Please align with latest changes in the dev branch @Abrom8

moritzkirstein (approved):

oceanByte (approved):

PR #182, opened 2022-08-12
feat: increase transparency using checkboxes disclaimers

### LINKED ISSUE #177: [Feature] Increase transparency using checkboxes & disclaimers
## Motivation / Problem
There is potential to increase transparency towards our users regarding the immutability of blockchains.

## Solution
(1) When publishing a service offering we can introduce (two) checkboxes next to the "SUBMIT" button which must be selected before the service offering can be published. Suggested text:
- I agree that my wallet address and public key will be transmitted to a smart contract stored permanently on the Gaia-X testnet, so I can prove service offering ownership and sales.
- I confirm that I did not provide personal data in the metadata, which will be stored permanently on-chain on the Gaia-X testnet.

(2) When consuming a service offering, we can add a disclaimer below the "BUY COMPUTE JOB" / "GET"button. Suggested text:
If you consume a service offering, your wallet address and public key will be stored permanently on-chain on the Gaia-X testnet. For more information, please refer to our [privacy policy](https://portal.minimal-gaia-x.eu/privacy/en).

## Alternatives
Instead of checkboxes we could also add an additional disclaimer.

--- 
**Tasks:**
- [x] Add required checkbox at end of publishing form: ```I agree that my wallet address and public key will be transmitted to a smart contract stored permanently on the Gaia-X testnet, so I can prove service offering ownership and sales.```
- [x] Add required checkbox at end of publishing form: ```I confirm that I did not provide personal data in the metadata, which will be stored permanently on-chain on the Gaia-X testnet.```
- [x] Add disclaimer in `<ButtonBuy>` component, below the button itself: ```
If you consume a service offering, your wallet address and public key will be stored permanently on-chain on the Gaia-X testnet. For more information, please refer to our [privacy policy](https://portal.minimal-gaia-x.eu/privacy/en).```

LucaMilanese90:
Fixes #177

## Proposed Changes

  - add checkboxes before publish
  - add disclaimer under asset buy button

### DIFF
--- a/src/components/atoms/ButtonBuy.tsx
+++ b/src/components/atoms/ButtonBuy.tsx
@@ -1,8 +1,35 @@
+import { graphql, useStaticQuery } from 'gatsby'
 import React, { FormEvent, ReactElement } from 'react'
 import Alert from './Alert'
 import Button from './Button'
 import styles from './ButtonBuy.module.css'
 import Loader from './Loader'
+
+const query = graphql`
+  query {
+    content: allFile(filter: { relativePath: { eq: "assetDisclaimer.json" } }) {
+      edges {
+        node {
+          childContentJson {
+            message
+          }
+        }
+      }
+    }
+  }
+`
+
+interface DisclaimerData {
+  content: {
+    edges: {
+      node: {
+        childContentJson: {
+          message: string
+        }
+      }
+    }[]
+  }
+}
 
 interface ButtonBuyProps {
   action: 'download' | 'compute'
@@ -147,6 +174,9 @@ export default function ButtonBuy({
   algorithmPriceType,
   algorithmConsumableStatus
 }: ButtonBuyProps): ReactElement {
+  const data: DisclaimerData = useStaticQuery(query)
+  const { message } = data.content.edges[0].node.childContentJson
+
   const buttonText =
     action === 'download'
       ? hasPreviousOrder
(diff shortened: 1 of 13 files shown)


src/components/atoms/ButtonBuy.tsx
```diff
@@ -215,6 +215,10 @@ export default function ButtonBuy({
                   selectedComputeAssetType,
                   algorithmConsumableStatus
                 )}
+            <Alert
+              text="If you consume a service offering, your wallet address and public key will be stored permanently on-chain on the Gaia-X testnet. For more information, please refer to our [privacy policy](/privacy/en)."
+              state="info"
+            />
```
moritzkirstein, line 221: question(blocking): Can we query this from our content folder? Would be cleaner and follow the repo standard for content.

moritzkirstein (changes requested):

moritzkirstein (approved):

moritzkirstein (changes requested):
Code looks good. Noticed a minor issue in the preview, though: the `walletAddress` checkbox is acting in the opposite way somehow. When checked the value in the DDO preview is "false"
grafik

LucaMilanese90:
Can you please try again deleting the previously saved values from local storage? Just to understand if that is the cause or if it's unrelated

moritzkirstein:
Yes it is the issue. Works as expected with clean local storage.

LucaMilanese90:
This might be a little problematic when we'll deploy. Maybe we should consider to put a time limit to decide if we want to use the values from local storage or show an empty form, does not make much sense to pre-fill a form with one week old data imho.

moritzkirstein:
> This might be a little problematic when we'll deploy. Maybe we should consider to put a time limit to decide if we want to use the values from local storage or show an empty form, does not make much sense to pre-fill a form with one week old data imho.

Not that big of a problem imo. For these values lets never use cached data. This is about user consenting they read terms / disclaimers, so we should request active input on every submit, anyways, if that makes sense.

LucaMilanese90:
@moritzkirstein I updated the form to now exclude all the `<Terms />` values from local storage. This should fix all the problems.

moritzkirstein (approved):

PR #184, opened 2022-08-16
feat: exclude "thing" type assets from query results

### LINKED ISSUE #183: [BUG] Asset details page broken
## Summary

Our asset details page crashes for certain assets: We introduced a new asset type that can currently not be handled by our portal. For now we want to simply remove these assets from the portal.
(Example: https://deploy-preview-173--mvg-portal-deltadao.netlify.app/asset/did:op:94E413D68c2f0c2DB39370C012521CC10a335eca)

## Solution
Adjust aquarius baseQuery to exclude assets of type 'thing' by default
```metadata.attributes.main.type = 'thing' ```

### PR DESCRIPTION
**Pull request by LucaMilanese90**
Fixes #183

## Proposed Changes

  - exclude "thing" type assets from query results

### DIFF
--- a/src/utils/aquarius.ts
+++ b/src/utils/aquarius.ts
@@ -69,10 +69,10 @@ export function getWhitelistShould(): // eslint-disable-next-line camelcase
 }
 
 export function getDynamicPricingMustNot(): // eslint-disable-next-line camelcase
-{ must_not: FilterTerm } | undefined {
+FilterTerm | undefined {
   return allowDynamicPricing === 'true'
     ? undefined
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/utils/aquarius.ts [hunk 1]
```diff
@@ -104,6 +112,7 @@ export function generateBaseQuery(
       SortDirectionOptions.Descending
   }
 
+  console.log(generatedQuery)
```
moritzkirstein, line 115: issue(blocking): should remove console.log here

moritzkirstein (changes requested):

moritzkirstein (approved):

PR #189, opened 2022-08-16
enhancement: replace wallet address checkbox with disclaimer

### LINKED ISSUE #188: [Enhancement] checkbox titles and disclaimer
I think it makes more sense to transform the first checkbox in a disclaimer below the submit button which states: "When you click submit, your wallet address and public key will be transmitted to a smart contract stored permanently on the Gaia-X testnet."
Regarding the second check-box, could we please change the heading to "Metadata confirmation"?

LucaMilanese90:
Fixes #188

## Proposed Changes

  - transform "wallet address" checkbox to disclaimer
  - update "Metadata confirmation" checkbox title

### DIFF
--- a/src/components/pages/Publish/FormActions.tsx
+++ b/src/components/pages/Publish/FormActions.tsx
@@ -1,40 +1,46 @@
 import React, { FormEvent, ReactElement } from 'react'
 import { useOcean } from '../../../providers/Ocean'
 import { useWeb3 } from '../../../providers/Web3'
+import Alert from '../../atoms/Alert'
 import Button from '../../atoms/Button'
 import styles from './FormActions.module.css'
 
 export default function FormActions({
   isValid,
-  resetFormAndClearStorage
+  resetFormAndClearStorage,
+  walletDisclaimer
 }: {
   isValid: boolean
   resetFormAndClearStorage: (e: FormEvent<Element>) => void
+  walletDisclaimer: string
 }): ReactElement {
   const { ocean, account } = useOcean()
   const { isChainIdAllowed } = useWeb3()
 
   return (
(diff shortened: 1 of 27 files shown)


src/components/pages/Publish/FormActions.tsx
```diff
@@ -1,40 +1,46 @@
...
         </Button>
-      )}
+
+        {status !== 'empty' && (
+          <Button style="text" size="small" onClick={resetFormAndClearStorage}>
+            Reset Form
+          </Button>
+        )}
+      </div>
+      <Alert state="info" text={walletDisclaimer} />
```
moritzkirstein, line 43: suggestion(non-blocking): I like that we use the `info` state here. However checking the preview it seems that warning & info use the same color scheme. Would expect something like our blue or gray for information, and use the orange for warning only.

I guess not high prio, only fix if this is small scope

moritzkirstein (approved):

src/components/pages/Publish/FormActions.tsx
LucaMilanese90: I agree, I'll check if it's used somewhere else and update it with the gray color scheme, should take just a few minutes.

LucaMilanese90: Update: I checked and the `info' state is used in a few other places. The idea would be to turn those places to `warning` (to keep the current color) and just keep info for the new disclaimer under the buy button, would it be ok for you?

moritzkirstein: Yes, sounds right! Let's update both new disclaimers to info scheme (publishing and consuming)

moritzkirstein (approved):

PR #193, opened 2022-08-22
Fix: aquarius id queries

### LINKED ISSUE #192: [BUG] Aquarius queries return no hits
## Summary

Featured Asset query seems to be broken

## Current Behavior

No featured assets are shown
  
## Expected Behavior

Featured assets are listed correctly

moritzkirstein:
Fixes #192 

## Proposed Changes

  - change aqua query to `_id` instead of `id`
  -
  -


LucaMilanese90 (approved):

oceanByte (approved):
It works, but it is still strange that we need to update this without any changes in the backend. We should make sure to investigate this topic. But we can merge this change to fix the issue now.

PR #205, opened 2022-08-31
enhancement: menu ux update

### LINKED ISSUE #201: [Enhancement] Menu & UX Update (WIP)
## Motivation / Problem

We want to enhance usability and the MVG menu / navigation.

## Solution 

Mock: https://www.figma.com/file/mgBcp6CwzOl1e4hNIfaBKV/MVG-Nav?node-id=1%3A2

- [x] Move statistics to footer
- [x] Add `Catalogue` to navigation -> links to searchpage
- [x] Update ecosystems dropdown to display relevant projects
- [x] Move Profile to wallet dropdown (account dropdown)
- [x] Include Debug Mode (settings) in account dropdown
- [x] Include link to or setting to turn on Onboarding in account dropdown
- [x] Remove settings button

LucaMilanese90:
Fixes #201

## Proposed Changes

  - Move statistics to footer
  - Add Catalogue to navigation -> links to searchpage
  - Update ecosystems dropdown to display relevant projects
  - Move Profile to wallet dropdown (account dropdown)
  - Include Debug Mode (settings) in account dropdown
  - Include link to or setting to turn on Onboarding in account dropdown
  - Remove settings button

### DIFF
--- a/src/components/molecules/Menu.tsx
+++ b/src/components/molecules/Menu.tsx
@@ -29,7 +29,7 @@ function MenuLink({ name, link }: { name: string; link: string }) {
 }
 
 export default function Menu(): ReactElement {
-  const { menu, badge } = useSiteMetadata()
+  const { menu } = useSiteMetadata()
 
   return (
     <div className={styles.wrapper}>
--- a/src/components/molecules/Menu.tsx
+++ b/src/components/molecules/Menu.tsx
@@ -1,5 +1,5 @@
 import React, { ReactElement } from 'react'
-import { Link, prefetchPathname } from 'gatsby'
+import { Link } from 'gatsby'
 import { useLocation } from '@reach/router'
 import loadable from '@loadable/component'
 import styles from './Menu.module.css'
--- a/src/components/molecules/Menu.tsx
+++ b/src/components/molecules/Menu.tsx
@@ -4,7 +4,6 @@ import { useLocation } from '@reach/router'
 import loadable from '@loadable/component'
 import styles from './Menu.module.css'
 import { useSiteMetadata } from '../../hooks/useSiteMetadata'
-import UserPreferences from './UserPreferences'
 import Logo from '../atoms/Logo'
 import Networks from './UserPreferences/Networks'
 import SearchBar from './SearchBar'
@@ -55,7 +54,6 @@ export default function Menu(): ReactElement {
             <SearchBar />
             <Networks />
             <Wallet />
-            <UserPreferences />
           </div>
         </nav>
       </Container>
--- a/src/components/molecules/Menu.tsx
+++ b/src/components/molecules/Menu.tsx
@@ -12,7 +12,13 @@ import SearchButton from './SearchButton'
 
 const Wallet = loadable(() => import('./Wallet'))
 
-function MenuLink({ name, link }: { name: string; link: string }) {
+export function MenuLink({
+  name,
+  link
+}: {
+  name: string
+  link: string
+}): ReactElement {
   const location = useLocation()
 
   const classes =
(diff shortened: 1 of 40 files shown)


moritzkirstein (changes requested):
- Log should be opened in a new tab
- Ecosystems should be opened in a new tab
- Statistics should be opened in a new tab
- Future Mobility should link to https://marketplace.future-mobility-alliance.org/

In general I am thinking if the `Log` entry makes sense in the top navigation or if this could/should be moved to the footer as well... We link to the exchangelog from every tsx anyways. Opinions?

LucaMilanese90:
It seems a little bit too much to have it on the menu imho, but it's also true that our current footer is not that discoverable (small and low contrast, easy to miss if you are not actively looking for it). Another idea would be to create a dedicated page with all the links we want to add, we would have the space to properly explain what the link is and it would result in a single link in the menu

moritzkirstein:
Okay. Could be an apporach, I just fear the overhead this creates. We can think about updating the footer to be a bit more discoverable - smth like we did with the feat/mvp branch. If we manage to reuse components this could be managable effort and could be done anytime (not too high priority for now).

LucaMilanese90:
Agreed, that footer was much better looking than the one we have right now for a relatively low implementation effort. But we should make sure everyone is fine with this, I remember Kai specifically mentioned that putting links in the footer is not a valid replacement for the the menubar, so before we move stuff around we should probably agree on the placement.

moritzkirstein (commented):
grafik

Not yet that happy with this section. It looks a bit off to me. Maybe we can move the profile to the top? Or have the account info first and the profile in third position.
Also the "VIEW PROFILE" link itself could be more present imo, maybe coloring it differently will already help.
Currently I fear users miss this section and won't be able to find this if they are not used to the Web3 account addresses. Maybe we need a "account" icon to communicate this better.

LucaMilanese90:
> The navbar is much more helpful now! Couple of suggestions:
> 
> - [ ]  Issue: External links should have some kind of indicator (like an icon) that they are pointing to an external resource
>  - [ ]  Issue: Hovering "Ecosystems" should automatically show the menu. I do not want to click it first.
>  - [ ]  Suggestion: Also, automatically show the menu when hovering over the account.
> - [ ] Suggestion: Remove the GX OCEAN section completely (they have to set up Metamask anyways)
> - [ ] Suggestion: "View Profile" should be the first entry in the account menu
> - [ ] Suggestion: Make "View Profile" the same dark blue color as the other links from the menu. Because it leads to a page and does not trigger a modal like other light blue links in the account menu

LucaMilanese90:
Would something like this be ok for signaling an external link? 
Screen Shot 2022-09-02 at 12 44 46

moritzkirstein:
In our privacy policy I think we have a symbol already for external links:
minimal-gaia-x.eu/privacy/en

LucaMilanese90:
> In our privacy policy I think we have a symbol already for external links: https://minimal-gaia-x.eu/privacy/en#6-external-links Maybe we can (or should) just reuse this?
> 
> **Edit** seems like this was removed already. In that case I am fine with the symbol above, before that we used a small arrow symbol to the right of these links

It's still there, I just found it and added to the menubar

moritzkirstein:
Could also be added to `Statistics` in the footer

src/components/molecules/Menu.tsx
```diff
@@ -1,35 +1,58 @@
 import React, { ReactElement } from 'react'
-import { Link } from 'gatsby'
+import { Link, prefetchPathname } from 'gatsby'
```
moritzkirstein, line 2: question: seems like `prefetchPathname` is not used. Can the import be removed?

moritzkirstein (approved):

PR #213, opened 2022-09-06
enhancement: new footer

### LINKED ISSUE #217: [Enhancement] New footer
## Motivation / Problem
The current footer is very simple and does not allow the inclusion of useful links.

## Solution 
Update the layout of the footer to include external links (Exchange log, statistics API page…), site navigation, and the deltaDao logo

LucaMilanese90:
Fixes #217 

## Proposed Changes

  - add new footer style


moritzkirstein (changes requested):
Okay very nice! I have some ideas / porposals:

- maybe we have the title of the portal down in the footer as well (logo position)
- move the logo to the bottom left corner
- we could include our slogan into the footer - somewhere close to the logo: "data economy solutions - GDPR compliant"
- I would like the `Bookmarks` link to be included in the wallet dropdown (below profile) -> maybe with the bookmarks icon used on AssetDetails
- I still think we can remove external links from the main navigation. The new footer allows for this - would move forward with this change and then get feedback from the team on the preview.

Maybe we can try some of these ideas and see what makes sense

moritzkirstein (approved):

PR #224, opened 2022-09-07
fix: default tab selection in profile page

### LINKED ISSUE #222: [BUG] The defaultTab parameter on the profile page has no effect
## Summary

The `defaultTab` parameter on `/profile` page has no effect
grafik
grafik

## Current Behavior

The tab defined in the parameter is not selected by default (on pafge load)

## Expected Behavior

The tab that is given to the profile page via the `defaultTab` parameter should be the one selected on page load

## Steps to Reproduce
1. Head to any CtD dataset
2. Start a compute job
4. Aftwer successful start, click on Go to History
5. Profile page should load with ComputeJobs Tab selected but the Download tab is selected

LucaMilanese90:
Fixes #222

## Proposed Changes

  - open profile page at the correct tab if specified in the url
  - show the "edge ctd device not available" warning only for edge assets

### DIFF
--- a/src/components/atoms/Tabs.tsx
+++ b/src/components/atoms/Tabs.tsx
@@ -9,19 +9,22 @@ interface TabsItem {
 
 export default function Tabs({
   items,
+  selectedIndex,
+  setSelectedIndex,
   className,
-  handleTabChange,
-  defaultIndex
+  handleTabChange
 }: {
   items: TabsItem[]
+  selectedIndex: number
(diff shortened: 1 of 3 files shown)


moritzkirstein (changes requested):
Accessing `/profile` page without defaultTab parameter present leads to a site crash - see console.error:

grafik

moritzkirstein (approved):

PR #233, opened 2022-09-13
fix: show history ctd items from different providers if they share computeAddress

### LINKED ISSUE #232: [BUG] ctd history tab shows duplicates and unsorted items
## Summary
the ctd history tab shows duplicates and unsorted items

## Expected Behavior
remove duplicates and sort the items

LucaMilanese90:
Fixes #232

## Proposed Changes

  - fetch ctd jobs from all the providers even if they share the same computeAddress
  - remove duplicate items in ctd history tab and sort by creation date

### DIFF
--- a/src/utils/compute.ts
+++ b/src/utils/compute.ts
@@ -140,16 +140,13 @@ async function getProviders(
       }
       const provider = await Provider.getInstance(instanceConfig)
       await provider.setBaseUrl(serviceEndpoints[i])
-      const hasSameCompute =
-        providers.filter((x) => x.computeAddress === provider.computeAddress)
-          .length > 0
-      if (!hasSameCompute) providers.push(provider)
+      providers.push(provider)
     }
   } catch (err) {
     Logger.error(err.message)
   }
 
-  return providers
+  return [...new Set(providers)]
 }
 
 async function getJobs(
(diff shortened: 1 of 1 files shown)


src/utils/compute.ts
```diff
@@ -203,7 +200,20 @@ async function getJobs(
...
+      uniqueAgreementIds.add(e.agreementId)
+      return isDuplicate
+    })
+    .sort(
+      (a, b) =>
+        new Date(Number(b.dateCreated) * 1000).getTime() -
+        new Date(Number(a.dateCreated) * 1000).getTime()
+    )
+
+  return uniqueComputeJobs
```
moritzkirstein, line 216: suggestion(non-blocking): refactor for readability - something like
```tsx
  const uniqueAgreementIds = new Set([...computeJobs.map(e => e.agreementId)])
  const uniqueComputeJobs = computeJobs.filter(e => !uniqueAgreementIds.has(e.agreementId))
  const sortedComputeJobs = uniqueComputeJobs.sort((a,b) =>
    // ...
  )

  return sortedComputeJobs
```

LucaMilanese90: Creating the set before filtering will not work because the items would be all incorrectly detected as "duplicates". But I can split the filter from the sort if you think it's more readable

moritzkirstein (approved):

PR #241, opened 2022-09-22
feat: edit "thing" assets metadata

LucaMilanese90:
## Proposed Changes

  - allow to edit metadata and service-SD of `thing` asset type


moritzkirstein:
Added don't merge label for now, as it is not a current priority.

PR #252, opened 2022-10-06
enhancement: rebrand MVG to Pontus-X

LucaMilanese90:
## Proposed Changes

  - replace MVG and minimal viable Gaia-X instances with the new Pontus-X - Gaia-X Web3 Ecosystem branding

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@
 
 ## 🏄 Get Started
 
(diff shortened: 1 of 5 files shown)


LucaMilanese90:
Should we also think about renaming the repository?

oceanByte (approved):

PR #274, opened 2022-12-07
feat: add checksum check for DIDs in /asset

LucaMilanese90:
## Proposed Changes

  - check if the did param is a valid eth address and navigate to checksum version

### DIFF
--- a/src/pages/asset/index.tsx
+++ b/src/pages/asset/index.tsx
@@ -4,13 +4,32 @@ import { PageProps } from 'gatsby'
 import PageTemplateAssetDetails from '../../components/templates/PageAssetDetails'
 import AssetProvider from '../../providers/Asset'
 import OceanProvider from '../../providers/Ocean'
+import Web3 from 'web3'
+import { useLocation, useNavigate } from '@reach/router'
 
 export default function PageGatsbyAssetDetails(props: PageProps): ReactElement {
+  const navigate = useNavigate()
+  const location = useLocation()
+  const { origin } = location
+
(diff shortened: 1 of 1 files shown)


src/pages/asset/index.tsx
```diff
@@ -4,13 +4,32 @@ import { PageProps } from 'gatsby'
...
+    if (isAddress && !isChecksumValid) {
+      const checksumAddress = Web3.utils.toChecksumAddress(address)
+      const checksumDid = `did:op:${checksumAddress.replace('0x', '')}`
+      setDid(checksumDid)
+      navigate(`${origin}/asset/${checksumDid}`)
+      return
+    }
+
+    setDid(did)
+  }, [location, navigate, origin, props.location.pathname])
```
moritzkirstein, line 32: suggestion/question(blocking): could we just simplify this and always calculate the checksum for the input? I feel like `checkAddressChecksum` and `toChecksumAddress` probably have a similar runtime anyways (where faster loadtime would be the only argument I see why we might wanna skip `toChecksumAddress` call)

moritzkirstein (changes requested):

src/pages/asset/index.tsx
LucaMilanese90: I was doing it on the first try but run into the issue of infinite redirects since every time the useEffect triggered and calculated the checksum. A simpler approach would be to just set the `did` and not redirect, but in this way the `did` in the url would not be correct.

moritzkirstein (approved):

PR #278, opened 2022-12-16
Feature: Polygon ID Verification Flow

moritzkirstein:
[MVP] Demonstrates Polygon ID verification flow

## Proposed Changes

  - Add `VerificationModal` on profile pages
  - Allow account owners to verify a Polygon ID claim based on ZKP
  - Extend `useProfile` hook to interact with verifier
  - Show `VerifiedPatch` on profile if member has previously proofed claim ownership


src/components/pages/Profile/Header/VerifiedPatch.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { ReactComponent as VerifiedPatchIcon } from '../../../../images/patch_check.svg'
+import { useProfile } from '../../../../providers/Profile'
+import Tooltip from '../../../atoms/Tooltip'
+import styles from './VerifiedPatch.module.css'
+
+export default function VerifiedPatch(): ReactElement {
+  const tooltip = (
+    <span>
+      This member has been officially onboarded to the GEN-X network.{' '}
+      <Link to="#">Learn more</Link>
```
LucaMilanese90, line 12: Question (blocking): it looks like there is a dead link here, where is the `Learn More` supposed to point to?

moritzkirstein: Good catch. I think we can remove this for now, unless we have a fitting section in the docs already @oceanByte @Abrom8

LucaMilanese90 (approved):

PR #280, opened 2023-01-09
feat: remove Gaia-X from available networks list

LucaMilanese90:


oceanByte (commented):
I suggest to remove all references to the old GX_NETWORK as it is not used anymore and there are no foreseeable events that would change this decision. We are also planning to shutdown the old network completely.

moritzkirstein (approved):

PR #287, opened 2023-01-27
refactor: remove unused dependencies

LucaMilanese90:
## Proposed Changes

  - remove datatoken price conversion to FIAT
  - remove purgatory checks for accounts and assets
  - remove ens deps


src/utils/purgatory.ts
```diff
@@ -0,0 +1,25 @@
...
+  return { did: undefined, reason: undefined }
+}
+
+export async function getAccountPurgatoryData(
+  address: string
+): Promise<PurgatoryDataAccount> {
+  // const data = await fetchData(`${purgatoryUrl}account?address=${address}`)
+  // return { address: data[0]?.address, reason: data[0]?.reason }
+  return { address: undefined, reason: undefined }
+}
```
moritzkirstein, line 25: thought(blocking): we should add a `/TODO` comment or similar mentioning why this was commented out

moritzkirstein (changes requested):

src/utils/purgatory.ts
LucaMilanese90: Alternatively, I was considering the possibility of adding the purgatory url to the `app.config` file and just adding a url check to the `purgatory.ts` file to evaluate if it should make the api call or not. Do you think it would be better or is it unnecessarly complex?

moritzkirstein: No also thought of this. If it does not take too much time go ahead and add this. Just make sure to comment in the config on how to use the variable (e.g., leave empty to skip purgatory calls or something like that)

LucaMilanese90: Should be better now, please have a look.

moritzkirstein (approved):

PR #295, opened 2023-02-17
feat: design update

LucaMilanese90:
## Proposed Changes

  - update design scheme
  - update GEN-X config


src/components/Footer/Links.tsx
```diff
@@ -1,49 +1,61 @@
...
+                </Link>
+              ) : (
+                <a
+                  key={i}
+                  href={e.link}
+                  target="_blank"
+                  rel="noopener noreferrer"
+                >
+                  {e.name} &#8599;
+                </a>
```
moritzkirstein, line 35: question: can't we use a `ExternalLink` component or smth similar here? I remember we had one in the past.

src/components/Header/Menu.tsx
```diff
@@ -2,57 +2,77 @@ import React, { ReactElement } from 'react'
...
+      {name}
     </Link>
+  ) : (
+    <a
+      href={link}
+      className={classes}
+      target="_blank"
+      rel="noopener noreferrer"
+    >
+      {name} &#8599;
```
moritzkirstein, line 46: question: see above. I remember some component to use here.

src/components/Home/index.tsx
```diff
@@ -59,25 +31,16 @@ export default function HomePage(): ReactElement {
...
-      <SectionQueryResult
-        title="Highest veOCEAN Allocations"
-        query={queryMostAllocation}
-      />
-
-      <SectionQueryResult title="Most Sales" query={queryMostSales} />
-      <MostViews />
-      <TopSales title="Publishers With Most Sales" />
-      <TopTags title="Top Tags By Sales" />
-
```
moritzkirstein, line 74: issue: I think some of these queries could actually also be interesting for us. I would suggest removing everything veOcean related (all "allocation" stuff) but keeping the other sections and going off of team/user feedback from there.

src/stylesGlobal/_colors.css
```diff
@@ -0,0 +1,42 @@
...
+
+  --ocean-grey-darker: #201f1f;
+  --ocean-grey-dark: #303030;
+  --ocean-grey: #41474e;
+  --ocean-grey-light: #8b98a9;
+  --ocean-grey-lighter: #e2e2e2;
+  --ocean-grey-dimmed: #f7f7f7;
+
+  --ocean-gradient: linear-gradient(to right bottom, #7b1173, #ff4092);
+}
```
moritzkirstein, line 42: I like the split of colors to a separate file.
suggestion: it would be nice to have a comment / section of required colors for the design to work. (e.g., there needs to be a `--brand-gradient` var, but `--brand-violet` currently seems optional / unused)

moritzkirstein (changes requested):
Nice overhaul. Looking forward to seeing this in action!
suggestion(non-blocking): for the `public/share.png` we should probably add another image. Not a priority for now. We could add a low prio issue on this, too.

LucaMilanese90:
I agree, I just wanted to remove the ocean logo but we can for sure come up with something new

src/components/@shared/FormInput/InputElement/FilesInput/Info.tsx
```diff
@@ -20,7 +20,7 @@ export default function FileInfo({
   return (
     <div className={`${styles.info}`}>
       <h3 className={`${styles.url} ${hideUrl ? styles.hideUrl : null}`}>
-        {hideUrl ? 'https://oceanprotocol/placeholder' : file.url}
+        {hideUrl ? 'https://delta-dao/the-future-is-now' : file.url}
```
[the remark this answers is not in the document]
LucaMilanese90: Sharing the credit for the idea with @Abrom8 during our last pairing session 😁

src/stylesGlobal/_colors.css
LucaMilanese90: The idea would be to replace the naming in the `_colors.css` file to reflect the actual colors, so our `--brand-blue` would be `--brand-red` in another portal, which is different from what we do in v3. This means that I would still need to manually go through the `_variables.css` file and input the colors where needed (especially to avoid putting together colors that don't work well combined). All things consider I don't think we can really identify "required" colors, at least until we make the whole process automated (but this would need some more refactoring, what we did is not enough for that IMHO).

moritzkirstein (approved):

PR #391, opened 2023-07-19
feat: service credentials id check

LucaMilanese90:
## Proposed Changes

  - verify if the service credential ID match with the associated asset's DID
  - rename service self-description into service credential


LucaMilanese90:
Example with valid credential and matching id: https://mvg-portal-git-feat-service-credentials-additio-e1f5bc-deltadao.vercel.app/asset/did:op:2e148b5c1ff12590c13bc018100743f0253f22f49cbaf7876612b2417f961789
Example with valid credential and non-matching id: https://mvg-portal-git-feat-service-credentials-additio-e1f5bc-deltadao.vercel.app/asset/did:op:bb0e9a01fda53c71e328eb7ccc91215c63a0964e9d3435875553e4754238b15e

src/components/@shared/VerifiedBadge/index.tsx
```diff
@@ -34,32 +53,26 @@ export default function VerifiedBadge({
...
-            <span className={styles.lastVerified}>
-              last check: <Time date={new Date().toString()} relative />
-            </span>
-          )}
+      {isLoading ? (
+        <Loader message="Verifying Service Credential" />
+      ) : (
+        <div className={styleClasses}>
+          <Badge isValid={isValid} label="Service Credential" />
+          <Badge isValid={idMatch} label="Credential ID match" />
```
moritzkirstein, line 61: suggestion: change label to something like "Verified Service", as this will most likely be a more elaborate check in the future (other than just ID=DID)

moritzkirstein (approved):

src/components/@shared/VerifiedBadge/index.tsx
kaimeinke: Also happy with that, but since only the credential is verified and not the service (today) I would not do it now.

PR #392, opened 2023-07-19
General enhancements

LucaMilanese90:
## Proposed Changes

  - move teaser network label to bottom right corner
  - update provider toast error message for accountId not in allow list
  - update provider toast error message for accountId in deny list
  - update access datatoken icon on publish flow


src/@utils/accessDetailsAndPricing.ts
```diff
@@ -191,6 +191,30 @@ export async function getOrderPriceAndFees(
   } catch (error) {
     const message = getErrorMessage(JSON.parse(error.message))
     LoggerInstance.error('[Initialize Provider] Error:', message)
+
+    // Customize error message for accountId non included in allow list
+    if (
+      message.includes(
+        'ConsumableCodes.CREDENTIAL_NOT_IN_ALLOW_LIST' || 'denied with code: 3'
+      )
+    ) {
```
LucaMilanese90, line 200: I had to look for a substring since the provider is only outputting a generic 400 error. Looking at the provider code, I think the correct error should output `denied with code:3`, but there might be an error with the enum since instead of the error code we are getting the enum itself `ConsumableCodes.CREDENTIAL_NOT_IN_ALLOW_LIST`.
See https://github.com/oceanprotocol/provider/blob/main/ocean_provider/utils/consumable.py and https://github.com/oceanprotocol/provider/blob/176564b67dd505dbdb0e4614794e24d50b283fb9/ocean_provider/utils/credentials.py#L43

moritzkirstein (approved):

moritzkirstein:
we should add a comment (or TODO) on the provider error message check, as this seems like it's unintended behaviour on provider end

PR #415, opened 2023-09-11
Features alignment

LucaMilanese90:
Fixes #

## Proposed Changes

  - https://github.com/oceanprotocol/market/pull/1958
  - https://github.com/oceanprotocol/market/pull/1921
  - align with `main` branch


src/@utils/docker.ts
```diff
@@ -16,13 +16,10 @@ export async function getContainerChecksum(
...
   }
   try {
-    const response = await axios.post(
-      `https://dockerhub-proxy.oceanprotocol.com`,
-      {
-        image,
-        tag
-      }
-    )
+    const response = await axios.post(`https://dockerhub-proxy.delta-dao.com`, {
```
moritzkirstein, line 19: suggestion: we should put this into a new appConfig variable

src/components/Profile/History/index.tsx
```diff
@@ -114,10 +115,12 @@ export default function HistoryPage({
     setRefetchJobs
   )
 
-  let defaultTabIndex = 0
-  defaultTab === 'ComputeJobs' ? (defaultTabIndex = 4) : (defaultTabIndex = 0)
```
moritzkirstein, line 118: suggestion: would be nice to retain this functionality (or re-enable it)
We may want to link to e.g., the compute history or the download tab and could achieve this by simply passing the tab index here for example

`.../profile?defaultTab=2` could link to compute history
```tsx
useEffect(() => {
  setTabIndex(defaultTab)
}, [defaultTab])
```
something like this

src/components/Profile/History/index.tsx
```diff
@@ -66,6 +66,7 @@ export default function HistoryPage({
   const [refetchJobs, setRefetchJobs] = useState(false)
   const [isLoadingJobs, setIsLoadingJobs] = useState(false)
   const [jobs, setJobs] = useState<ComputeJobMetaData[]>([])
+  const [tabIndex, setTabIndex] = useState(0)
```
moritzkirstein, line 69: suggestion: see below, we may want to initialize based on query param

moritzkirstein (approved):

src/components/Profile/History/index.tsx
LucaMilanese90: Implemented in https://github.com/deltaDAO/mvg-portal/pull/416

src/components/Profile/History/index.tsx
LucaMilanese90: Implemented in https://github.com/deltaDAO/mvg-portal/pull/416

PR #418, opened 2023-09-15
(WIP) Feat: Automated Wallet Interactions

moritzkirstein:
## Proposed Changes

  - introduce local automation
  -
  -


src/@context/Automation/AutomationProvider.tsx
```diff
@@ -0,0 +1,239 @@
...
+  updateBalance: () => Promise<void>
+  setIsAutomationEnabled: (isEnabled: boolean) => void
+  deleteCurrentAutomationWallet: () => void
+  importAutomationWallet: (encryptedJson: string) => Promise<boolean>
+  hasValidEncryptedWallet: () => boolean
+  decryptAutomationWallet: (password: string) => Promise<boolean>
+}
+
+// Refresh interval for balance retrieve - 20 sec
+const refreshInterval = 20000
```
oceanByte, line 42: thought: The `refreshInterval` is set to 20 seconds. Depending on the number of users and the scale of the application, this might result in a lot of requests. We should consider this in our architecture planning (focus on the nodes) as our user base grows.

src/@context/Automation/AutomationProvider.tsx
```diff
@@ -0,0 +1,239 @@
...
+    try {
+      if (balanceNativeToken)
+        setNativeBalance({
+          symbol: balanceNativeToken?.symbol.toLowerCase() || 'ETH',
+          balance: balanceNativeToken?.formatted
+        })
+
+      if (approvedBaseTokens?.length > 0) {
+        const newBalance: UserBalance = {}
+        await Promise.all(
```
oceanByte, line 105: suggestion: While using  `await Promise.all(...)`  efficient, if one of the promises fails, it will reject the entire batch. If we want to handle errors for each token individually and continue fetching balances for the others, we should consider using something like `Promise.allSettled()`.

src/components/Header/UserPreferences/Automation/Decrypt.tsx
```diff
@@ -0,0 +1,57 @@
...
+      {isLoading ? (
+        <Loader message="Decrypting..." />
+      ) : (
+        <>
+          <strong className={styles.warning}>The wallet is locked!</strong>
+          <form
+            onSubmit={async (e) => {
+              e.preventDefault()
+              decrpytToastRef.current = toast.info(`Decrypting Wallet...`)
+              await decryptAutomationWallet(passwordInputRef.current.value)
```
oceanByte, line 35: issue: The password is directly accessed from the passwordInputRef without any validation. This can be a potential security risk.

suggestion: Before attempting to decrypt the wallet, validate the password input. Ensure it's not empty and meets any required criteria (e.g., minimum length, contains special characters, etc.).

src/components/Header/UserPreferences/Automation/Decrypt.tsx
```diff
@@ -0,0 +1,57 @@
...
+    decryptAutomationWallet,
+    setIsAutomationEnabled
+  } = useAutomation()
+
+  const decrpytToastRef = useRef(null)
+  const passwordInputRef = useRef(null)
+
+  useEffect(() => {
+    toast.update(decrpytToastRef.current, { progress: decryptPercentage })
+  }, [decrpytToastRef, decryptPercentage])
```
oceanByte, line 22: nitpick: The variable name decrpytToastRef seems to have a typo. It might be clearer if named decryptToastRef.

src/components/Header/UserPreferences/Automation/Decrypt.tsx
```diff
@@ -0,0 +1,57 @@
...
+      ) : (
+        <>
+          <strong className={styles.warning}>The wallet is locked!</strong>
+          <form
+            onSubmit={async (e) => {
+              e.preventDefault()
+              decrpytToastRef.current = toast.info(`Decrypting Wallet...`)
+              await decryptAutomationWallet(passwordInputRef.current.value)
+              setIsAutomationEnabled(true)
+              toast.done(decrpytToastRef.current)
```
oceanByte, line 37: question: What happens if the decryptAutomationWallet function fails? It seems the toast will still indicate success with toast.done.

oceanByte: suggestion: Add error handling for the decryption process. If the decryptAutomationWallet function fails, provide feedback to the user, possibly using a toast error message. Or we just rely on the handling by the automation?

src/components/Header/UserPreferences/Automation/Details.tsx
```diff
@@ -0,0 +1,136 @@
...
+import styles from './Details.module.css'
+import Import from './Import'
+import Address from './Address'
+import Decrypt from './Decrypt'
+
+function AdvancedView(): ReactElement {
+  const { deleteCurrentAutomationWallet } = useAutomation()
+  const deleteWallet = () => {
+    deleteCurrentAutomationWallet()
+  }
```
oceanByte, line 18: question: You could directly use deleteCurrentAutomationWallet in the onClick handler without the need for the extra function. Why do we do it this way?

src/components/Header/UserPreferences/Automation/Import.tsx
```diff
@@ -0,0 +1,60 @@
...
+      const file = (target as any).files[0]
+      const reader = new FileReader()
+      reader.readAsText(file)
+      reader.onload = async (event) => {
+        await importAutomationWallet(event.target.result.toString())
+      }
+    } catch (e) {
+      LoggerInstance.error(e)
+    }
+  }
```
oceanByte, line 25: issue: The function directly reads the file and attempts to import it without any validation. This can be a potential security risk.

suggestion: Before importing the wallet from the JSON file, validate its content. Ensure it's not empty and meets any required criteria (e.g., has specific properties that a wallet JSON should have).

src/components/Header/UserPreferences/Automation/Import.tsx
```diff
@@ -0,0 +1,60 @@
...
+import styles from './Import.module.css'
+import Input from '../../../@shared/FormInput'
+import { LoggerInstance } from '@oceanprotocol/lib'
+
+export default function Import(): ReactElement {
+  const { isLoading, importAutomationWallet } = useAutomation()
+
+  const [showFileInput, setShowFileInput] = useState<boolean>()
+
+  const importToLocalStorage = async (target: EventTarget) => {
```
oceanByte, line 14: nitpick: The function name importToLocalStorage might be misleading as the function doesn't seem to import anything to local storage (the called function might do it, but this is not the responsibility of this function). Consider renaming it to better reflect its purpose, like importWalletFromFile.

src/components/Header/UserPreferences/Automation/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+    balance,
+    nativeBalance
+  } = useAutomation()
+
+  const [hasError, setHasError] = useState<boolean>()
+  const [hasWarning, setHasWarning] = useState<boolean>(false)
+
+  useEffect(() => {
+    balance &&
+      setHasWarning(
```
oceanByte, line 28: suggestion: The `hasWarning` state is set based on the balance of individual tokens being less than or equal to 0. It might be more intuitive to rename this state to something like `hasZeroTokenBalance` for clarity.

src/components/Header/UserPreferences/Automation/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  return (
+    <Tooltip
+      content={<Details isFunded={!hasError} />}
+      trigger="focus mouseenter click"
+      placement="bottom"
+      className={`${stylesIndex.preferences} ${wrapperClasses}`}
+    >
+      <div className={styles.wrapper}>
+        {!autoWallet && hasValidEncryptedWallet() && (
```
oceanByte, line 55: nitpick: The hasValidEncryptedWallet function is used as a condition for rendering the Lock icon. It might be clearer to store its result in a state or a variable for better readability.

oceanByte (changes requested):

app.config.js
```diff
@@ -83,6 +83,12 @@ module.exports = {
     process.env.NEXT_PUBLIC_DOCKER_HUB_PROXY_URL ||
     'https://dockerhub-proxy.delta-dao.com',
 
+  automationConfig: {
+    networkTokenFundDefaultValue: '2',
+    erc20ApprovalDefaultValue: '50',
+    roughTxGasEstimate: 0.02
```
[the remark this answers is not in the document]
moritzkirstein: agreed. This was meant to be a quick solution to provide some reference in the UI.
We should keep this in mind for a future enhancement I think.

src/@context/Automation/AutomationProvider.tsx
```diff
@@ -0,0 +1,239 @@
...
+      ethers.utils.isAddress(ethers.utils.getJsonWalletAddress(encryptedJson))
+    ) {
+      setAutomationWalletJSON(encryptedJson)
+      return true
+    } else {
+      toast.error('Could not import Wallet. JSON format invalid.')
+      LoggerInstance.error('Could not import Wallet. JSON format invalid.')
+      return false
+    }
+  }
```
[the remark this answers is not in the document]
moritzkirstein: I am not sure if this is already accessible at this point (while encrypted). If yes, then I aggree that this would be a great enhancement.

src/@context/Automation/AutomationProvider.tsx
moritzkirstein: Agreed. We also have another balance call on the portal for the connected wallet (e.g., Metamask) where we also use a interval (20s iirc)

src/components/Header/UserPreferences/Automation/Decrypt.tsx
moritzkirstein: Agreed on the requirements. However, the password is set by the user outside of our application. We don't really know the parameters users used for their password during encryption.

src/components/Header/UserPreferences/Automation/Decrypt.tsx
moritzkirstein: The `toast.done` only manages the state of the "progress" toast message. With done we will close this message in all states (failure & success)

The error or success states are handled separately in the [`decryptAutomationWallet` callback within the `AutomationProvider`](https://github.com/deltaDAO/mvg-portal/blob/bd1091bbd03677a39de9ed6a053cc15e0612acd7/src/%40context/Automation/AutomationProvider.tsx#L188-L195)

src/components/Header/UserPreferences/Automation/Details.tsx
moritzkirstein: Good catch. Agreed. Seems to be leftover from a previous iteration.

src/components/Header/UserPreferences/Automation/index.tsx
moritzkirstein: Thanks, agreed. Having this be a callback also seems to be a leftover from previous iterations. I refactored this to be a state variable instead.

src/components/Header/UserPreferences/Automation/Decrypt.tsx
oceanByte: Understood. But could we assume that it is always non-empty?

Abrom8 (changes requested):

src/components/Asset/AssetActions/Compute/index.tsx
```diff
@@ -58,6 +58,8 @@ import useNetworkMetadata from '@hooks/useNetworkMetadata'
 import { useAsset } from '@context/Asset'
 import WhitelistIndicator from './WhitelistIndicator'
 import { parseConsumerParameterValues } from '../ConsumerParameters'
+import { useAutomation } from '../../../../@context/Automation/AutomationProvider'
+import { Signer } from 'ethers'
 
 const refreshInterval = 10000 // 10 sec.
```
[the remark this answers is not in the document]
moritzkirstein: I reslve this, as it was not touched in this PR. Agreed in general.

src/components/Header/UserPreferences/Automation/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  const wrapperClasses = cx({
+    automation: true,
+    enabled: isAutomationEnabled
+  })
+
+  const indicatorClasses = cx({
+    indicator: true,
+    enabled: isAutomationEnabled,
+    warning: hasZeroERC20TokenBalances,
```
LucaMilanese90, line 45: thought (non-blocking): green and red are pretty clear indicators, but knowing that yellow indicates a zero balance of ocean and EUROe tokens might not be as clear. We could think about adding a warning or some sort of message to give the user a clue of what's happening.

LucaMilanese90 (changes requested):

LucaMilanese90:
Suggestion for a future update: the tooltip closes between the wallet json file import and the password input prompt. It might not be immediately apparent to the user that a second action is required. Can we have the tooltip to stay open throughout the whole operation?

src/@context/UserPreferences.tsx
```diff
@@ -77,6 +81,14 @@ function UserPreferencesProvider({
     localStorage?.allowExternalContent || false
   )
 
+  const [automationWallet, setAutomationWallet] = useState<string>(
+    localStorage?.automationWalletJSON || ''
+  )
+
+  const [automationWalletMode, setAutomationWalletMode] = useState<string>(
+    localStorage?.automationWalletMode || 'advanced'
```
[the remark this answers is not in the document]
moritzkirstein: Only personal preference so far. I think the earlier iteration of the advanced mode was a bit better suited for demos

Maybe we should just move this to appConfig and allow configuration of this.

moritzkirstein:
> Suggestion for a future update: the tooltip closes between the wallet json file import and the password input prompt. It might not be immediately apparent to the user that a second action is required. Can we have the tooltip to stay open throughout the whole operation?

This was not happening for me, but it is a good hint. I'll create an enhancement issue to keep track

LucaMilanese90:
> > Suggestion for a future update: the tooltip closes between the wallet json file import and the password input prompt. It might not be immediately apparent to the user that a second action is required. Can we have the tooltip to stay open throughout the whole operation?
> 
> This was not happening for me, but it is a good hint. I'll create an enhancement issue to keep track

Not sure why it happens to me, maybe it's a browser thing? 

https://github.com/deltaDAO/mvg-portal/assets/20192135/05ca5ac9-3732-455c-ae13-5fdbc4fce7ce

src/@context/UserPreferences.tsx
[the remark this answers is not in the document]
LucaMilanese90: I pointed this out because on the SBB portal we updated the onboarding module to show the simple view, so I thought it would make sense to have that as default to avoid confusing the user.

src/components/Header/UserPreferences/Automation/index.tsx
moritzkirstein: I created an issue for this 👍

src/@context/UserPreferences.tsx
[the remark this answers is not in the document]
moritzkirstein: Simply changed it to `SIMPLE` for now and added an issue to track the enhancement #440

src/components/Asset/AssetActions/Compute/index.tsx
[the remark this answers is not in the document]
moritzkirstein: yes that's correct. This was considered an additional feature/enhancement which needs to be tackled next. I'll use the comment to create an issue

src/components/Header/UserPreferences/Automation/Decrypt.tsx
```diff
@@ -0,0 +1,57 @@
...
+        <Loader message="Decrypting..." />
+      ) : (
+        <>
+          <strong className={styles.warning}>The wallet is locked!</strong>
+          <form
+            onSubmit={async (e) => {
+              e.preventDefault()
+              decryptToastRef.current = toast.info(`Decrypting Wallet...`)
+              await decryptAutomationWallet(passwordInputRef.current.value)
+              setIsAutomationEnabled(true)
```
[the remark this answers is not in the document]
moritzkirstein: good catch, added the missing if statement

src/components/Header/UserPreferences/Automation/Decrypt.tsx
moritzkirstein: I think it makes sense. Added a minimal check

LucaMilanese90 (approved):

oceanByte (approved):

PR #421, opened 2023-09-27
feat: update filters

LucaMilanese90:
## Proposed Changes

  - update filter and sort logic
  - update filter and sort UI at `/search` and `/profile`


src/components/Search/Filter.tsx
```diff
@@ -0,0 +1,298 @@
...
+  },
+  {
+    id: 'accessType',
+    label: 'Access Type',
+    type: 'filterList',
+    options: [
+      { label: 'download', value: FilterByAccessOptions.Download },
+      { label: 'compute', value: FilterByAccessOptions.Compute }
+    ]
+  },
```
moritzkirstein, line 46: suggestion: keeping these here makes sense imo, as we probably want these as default filters for every portal (see below)

src/components/Search/Filter.tsx
```diff
@@ -0,0 +1,298 @@
...
+      {
+        label: 'text analysis',
+        value: 'textAnalysis'
+      },
+      {
+        label: 'finance',
+        value: 'finance'
+      }
+    ]
+  }
```
moritzkirstein, line 69: suggestion: for custom filters I would suggest setting up a config file, similar to the address.config file
e.g., `filters.config.js`
Would be easy to quickly change / setup filters for every portal that way

src/components/Search/Filter.tsx
```diff
@@ -0,0 +1,298 @@
...
+  ],
+  manufacturing: [
+    'euprogigant',
+    'industry40',
+    'manufacturing',
+    'predictive-maintenance'
+  ],
+  textAnalysis: ['library', 'ocr', 'text-analysis'],
+  finance: ['graphql']
+}
```
moritzkirstein, line 90: suggestion: the same applies for these set definitions

maybe a config file just to specify filter-sets makes sense at that point (`const filterSets` and `id: 'filterSet'` could be moved there)

src/components/Search/Filter.tsx
```diff
@@ -0,0 +1,298 @@
...
+    if (addFiltersToUrl) {
+      let urlLocation = await addExistingParamsToUrl(location, [filterId])
+
+      if (filter.length > 0 && urlLocation.indexOf(filterId) === -1) {
+        const parsedFilter = filter.join(',')
+        urlLocation = `${urlLocation}&${filterId}=${parsedFilter}`
+      }
+
+      router.push(urlLocation)
+    }
```
moritzkirstein, line 150: suggestion: we can use early return pattern here

src/components/Search/Filter.tsx
```diff
@@ -0,0 +1,298 @@
...
+
+  async function applyClearFilter(addFiltersToUrl: boolean) {
+    const clearedFilters = { ...filters }
+    Object.keys(clearedFilters).forEach((key) => (clearedFilters[key] = []))
+    setFilters(clearedFilters)
+
+    if (ignorePurgatory !== undefined && setIgnorePurgatory !== undefined)
+      setIgnorePurgatory(true)
+
+    if (addFiltersToUrl) {
```
moritzkirstein, line 170: suggestion: early return could be used here, too

src/components/Search/Filter.tsx
```diff
@@ -0,0 +1,298 @@
...
+  async function handleSelectedFilter(value: string, filterId: keyof Filters) {
+    const updatedFilters = filters[filterId].includes(value)
+      ? { ...filters, [filterId]: filters[filterId].filter((e) => e !== value) }
+      : { ...filters, [filterId]: [...filters[filterId], value] }
+    setFilters(updatedFilters)
+
+    await applyFilter(updatedFilters[filterId], filterId)
+  }
+
+  async function applyClearFilter(addFiltersToUrl: boolean) {
```
moritzkirstein, line 162: suggestion: just naming this `clearFilters` is maybe a bit easier for readability.

src/components/Search/sort.tsx
```diff
@@ -18,69 +18,133 @@ const sortItems = [
   { display: 'Price', value: SortTermOptions.Price }
 ]
 
+const sortDirections = [
+  { display: '\u2191 Asc', value: SortDirectionOptions.Ascending },
+  { display: '\u2193 Desc', value: SortDirectionOptions.Descending }
```
moritzkirstein, line 23: suggestion: I'd prefer spelling out "Ascending" / "Descending" respectively

moritzkirstein (changes requested):

src/@hooks/useDebounce.tsx
```diff
@@ -0,0 +1,12 @@
+import { useEffect, useCallback } from 'react'
+
+const useDebounce = (effect, dependencies: any[], delay: number) => {
+  const callback = useCallback(effect, dependencies)
```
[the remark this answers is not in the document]
LucaMilanese90: I also tried this in the first iteration but it was causing an infinite loop so I took it out. Not really sure why it happens tbh.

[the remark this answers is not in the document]
oceanByte: Fine for now, but we should circle back to this when @moritzkirstein returns to address it properly.

oceanByte (approved):

src/@hooks/useDebounce.tsx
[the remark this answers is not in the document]
LucaMilanese90: I'll open an issue

oceanByte (approved):

PR #433, opened 2023-10-11
feat: add dev preview banner

LucaMilanese90:
## Proposed Changes

  - add alert banner for portal preview deployment

### DIFF
--- a/content/site.json
+++ b/content/site.json
@@ -68,6 +68,7 @@
     }
   ],
   "announcement": "",
+  "devPreviewAnnouncement": "You are visiting the Pontus-X Developer Preview. Please note that some functionalities are in beta and may not work as expected.",
   "warning": {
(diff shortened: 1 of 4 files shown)


content/site.json
```diff
@@ -68,6 +68,7 @@
     }
   ],
   "announcement": "",
+  "devPreviewAnnouncement": "You are visiting the Pontus-X Developer Preview. Please note that some functionalities are in beta and may not work as expected.",
```
moritzkirstein, line 71: suggestion: I think this supports markdown, so we could add a link to main version:

```suggestion
  "devPreviewAnnouncement": "You are visiting the Pontus-X Developer Preview. Please note that some functionalities are in beta and may not work as expected. Visit the [Pontus-X portal](https://pontus-x.eu) instead.",
```

If it does not support md, we can skip this

moritzkirstein (approved):

PR #467, opened 2023-11-05
fix: ctd algo pre-selection 

LucaMilanese90:
## Proposed Changes

  - only pre-select algo in c2d form if user is allowed to consume


src/components/Asset/AssetActions/Compute/FormComputeDataset.tsx
```diff
@@ -120,14 +120,18 @@ export default function FormStartCompute({
     })
     return assetDdo
   }
-
+  console.log(algorithms)
```
moritzkirstein, line 123: issue: should be removed or changed to Logger

moritzkirstein (approved):

PR #507, opened 2024-02-20
Feat: dedicated json file for verified wallets

moritzkirstein:
## Proposed Changes

  - create JSON file with wallet-address <> business mappings


address.config.js
```diff
@@ -1,3 +1,5 @@
+const PONTUSX_WALLETS = require('./pontusxWallets.json')
```
oceanByte, line 1: issue(blocking): We are not referencing any wallets, but addresses. I know that we used to name the property "verifiedWallets" (which is incorrect) but we should be precise with our wording here or might confuse other developers. 

We should update it to something like `const PONTUSX_ADDRESSES = require('./pontusxAddresses.json')`. And accordingly all references to the old property name must be updated as well.

oceanByte (changes requested):

address.config.js
moritzkirstein: Agreed, just didnt see it in scope initially. If it's blocking I'll update, should be quick enough

oceanByte (approved):

PR #521, opened 2024-03-11
Fix: wrong token being used for price button on Pontus-X

moritzkirstein:
Fixes an issue where the native token get's used for `<ButtonBuy>` _disabled_ prop, rather than the wrapped contract.

## Proposed Changes

  - Update `UserBalance` interface, splitting between native and approved balances


Abrom8 (approved):
lgtm

PR #524, opened 2024-03-21
fix: add wait to txs

Abrom8:
Fixes failed edit transactions. 

- adds wait to edit transactions 
- replace unresponsive subgraph endpoint

### DIFF
--- a/src/@utils/dispenser.ts
+++ b/src/@utils/dispenser.ts
@@ -14,12 +14,14 @@ export async function setMinterToPublisher(
     accountId,
     accountId
   )
-  await removeMinterTx.wait()
 
   if (!removeMinterTx) {
     setError('Updating DDO failed.')
     LoggerInstance.error('Failed at cancelMinter')
   }
+
+  await removeMinterTx.wait()
+
   return removeMinterTx
 }
 
@@ -40,5 +42,7 @@ export async function setMinterToDispenser(
     setError('Updating DDO failed.')
     LoggerInstance.error('Failed at makeMinter')
   }
+  await addMinterTx.wait()
+
   return addMinterTx
 }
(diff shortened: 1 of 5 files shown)


src/@utils/dispenser.ts
```diff
@@ -9,17 +9,18 @@ export async function setMinterToPublisher(
 ): Promise<ethers.providers.TransactionResponse> {
   const datatokenInstance = new Datatoken(signer)
 
-  const response = await datatokenInstance.removeMinter(
+  const removeMinterTx = await datatokenInstance.removeMinter(
     datatokenAddress,
     accountId,
     accountId
   )
+  await removeMinterTx.wait()
```
moritzkirstein, line 17: issue: `.wait()` could be called on undefined in case of tx failure

src/@utils/dispenser.ts
```diff
@@ -9,17 +9,18 @@ export async function setMinterToPublisher(
...
 
-  if (!response) {
+  if (!removeMinterTx) {
```
moritzkirstein, line 19: suggestion: move null check before .wait call

src/@utils/dispenser.ts
```diff
@@ -30,14 +31,14 @@ export async function setMinterToDispenser(
...
     accountId,
     accountId
   )
-  if (!response) {
+  if (!addMinterTx) {
     setError('Updating DDO failed.')
     LoggerInstance.error('Failed at makeMinter')
   }
-  return response
+  return addMinterTx
```
moritzkirstein, line 43: question: do we not need a .wait call here as well?

moritzkirstein (changes requested):

PR #575, opened 2024-09-10
Include FIWARE Portal

kaimeinke:
Including the FIWARE Portal for the FIWARE Global Summit


kaimeinke:
Blocking: SVG does not render correctly and needs to be changed.

oceanByte (approved):

PR #593, opened 2024-10-02
Feat/ddo export

KevinRohlf:
## Proposed Changes

  - add download DDO Button

### DIFF
--- a/src/@utils/downloadJSON.ts
+++ b/src/@utils/downloadJSON.ts
@@ -1,0 +1,23 @@
+export function isValidJSON(json: any): boolean {
+  try {
+    JSON.parse(json)
+    return true
+  } catch (e) {
+    return false
+  }
+}
+
+// Download JSON file
+export function downloadJSON(json: any, filename: string): void {
+  // check if the json is valid
+  if (isValidJSON(JSON.stringify(json))) {
+    const element = document.createElement('a') // create an <a> tag
+    const file = new Blob([JSON.stringify(json)], {
+      type: 'application/json'
+    }) // create a file
+    element.href = URL.createObjectURL(file) // create a URL for the file
+    element.download = `${filename}.json` // Set the file name
+    document.body.appendChild(element) // Append the tag to the body
+    element.click() // Click the a element
+  }
+}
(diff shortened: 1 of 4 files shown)


src/components/@shared/DDODownloadButton/index.tsx
```diff
@@ -0,0 +1,22 @@
+import { ReactElement } from 'react'
+import Button from '../atoms/Button'
+
+export default function DDODownloadButton({
+  asset
+}: {
+  asset: any
+}): ReactElement {
+  // Download DDO as JSON
+  const downloadJson = () => {
```
moritzkirstein, line 10: suggestion(blocking): it could make sense to check the `downloadFileBrowser` function that is exported by the ocean.js library

If that function is not working for our use-case, we should separate the `downloadJSON` function to be re-usable utility in the repository that accepts any JSON valid object and downloads it as a file.

src/components/@shared/DebugOutput/index.tsx
```diff
@@ -9,7 +10,16 @@ export default function DebugOutput({
...
-      {title && <h5>{title}</h5>}
+      <div
+        style={{
+          display: 'flex',
+          justifyContent: 'space-between',
+          flexWrap: 'wrap'
+        }}
+      >
+        {title && <h5>{title}</h5>}
+        {output && <DDODownloadButton asset={output} />}
```
moritzkirstein, line 21: issue(blocking): the download button should be available, even if the user is not in Debug-Mode

moritzkirstein (changes requested):

src/@utils/downloadJSON.ts
```diff
@@ -0,0 +1,23 @@
...
+    return true
+  } catch (e) {
+    return false
+  }
+}
+
+// Download JSON file
+export function downloadJSON(json: any, filename: string): void {
+  // check if the json is valid
+  if (isValidJSON(JSON.stringify(json))) {
```
moritzkirstein, line 13: issue(blocking): `JSON.stringify()` will always produce valid JSON, or run into an error itself. In this case the function should expect a JSON string as input rather than `any` object.

moritzkirstein (changes requested):

moritzkirstein (approved):

PR #598, opened 2024-10-10
feat: resolved address to display name

KevinRohlf:
## Proposed Changes

  - the names are now displayed in the asset header and in the asset teaser instead of the address

### DIFF
--- a/src/@utils/wallet/index.ts
+++ b/src/@utils/wallet/index.ts
@@ -53,8 +53,8 @@ export const connectKitTheme = {
 
 export function accountTruncate(
   account: string,
-  begin: number = 6,
-  end: number = 38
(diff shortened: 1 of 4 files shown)


src/@utils/wallet/index.ts
```diff
@@ -51,9 +51,13 @@ export const connectKitTheme = {
   '--ck-secondary-button-border-radius': 'var(--border-radius)'
 }
 
-export function accountTruncate(account: string): string {
+export function accountTruncate(
+  account: string,
+  begin: number = 6,
+  end: number = 38
```
oceanByte, line 57: nitpick(non-blocking): Could you add a comment explaining these default magic numbers as well?

oceanByte (approved):

oceanByte (approved):

PR #606, opened 2024-10-15
Fix/scrollbar position on publish flow

rawao:
Fixes #
- Fix scrollbar position on publish flow

### DIFF
--- a/src/components/Publish/Navigation/index.module.css
+++ b/src/components/Publish/Navigation/index.module.css
@@ -16,14 +16,6 @@
   }
 }
 
-@media (min-width: 60rem) {
-  .navigation {
-    padding: var(--spacer) 0;
-    margin-left: calc(var(--spacer) * -4);
-    margin-right: calc(var(--spacer) * -4);
-  }
-}
-
 .navigation ol {
   white-space: nowrap;
   text-align: center;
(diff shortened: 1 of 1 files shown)


moritzkirstein:
Can you please align with latest main branch @KevinRohlf ?

src/components/Publish/Navigation/index.module.css
```diff
@@ -1,10 +1,12 @@
 .navigation {
   margin: -1.3rem calc(var(--spacer) / -1.5) var(--spacer)
     calc(var(--spacer) / -1.5);
-  padding: calc(var(--spacer) / 2) 0;
   scrollbar-width: none;
   border-bottom: 1px solid var(--border-color);
   background-color: var(--background-highlight);
+  width: calc(100% + var(--spacer) 0 * 2);
```
moritzkirstein, line 7: issue(blocking): Is this needed for the functionality? width > 100% does not look correct to me. Also the math at the end looks weird (`0 * 2`).

suggestion: remove width here or simply keep at 100%

moritzkirstein (changes requested):

moritzkirstein (changes requested):
we introduce more space on big screens, rather than on small screens compared to the current version, see screenshots below:

**current:**
image

**new:**
image


I would suggest to, either:
1. revoke this change
2. introduce the additional space on mobile / small screens rather than bigger screens

moritzkirstein (approved):

PR #611, opened 2024-10-23
fix: update footer subtitle width

KevinRohlf:
Fixes #
- Footer logo subtitle made responsive

### DIFF
--- a/src/components/Footer/Footer.module.css
+++ b/src/components/Footer/Footer.module.css
@@ -41,13 +41,13 @@
   position: absolute;
   top: 4rem;
   left: 5rem;
-  width: max-content;
   max-width: 100%;
-  word-wrap: break-word;
 }
+
 @media screen and (min-width: 60rem) {
   .subtitle {
-    max-width: max-content;
+    max-width: unset;
+    width: max-content;
   }
 
   .main {
--- a/src/components/Footer/Footer.module.css
+++ b/src/components/Footer/Footer.module.css
@@ -27,7 +27,7 @@
   position: relative;
   text-align: left;
   color: var(--footer-font-color);
-  margin-bottom: 5vw;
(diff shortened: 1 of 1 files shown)


src/components/Footer/Footer.module.css
```diff
@@ -38,9 +39,20 @@
...
+  word-wrap: break-word;
+}
+@media screen and (min-width: 40rem) {
+  .subtitle {
+    max-width: 40vw;
+  }
+
+  .main {
+    margin-bottom: unset;
+  }
```
moritzkirstein, line 55: suggestion(blocking): I would suggest trying to utilize `100%` here rather than linking this to viewport.
We should also be able to utilize more space on mobile that way, as the navigation gets broken down into a new row in that view

moritzkirstein (changes requested):

src/components/Footer/Footer.module.css
moritzkirstein: example from preview: 
image

src/components/Footer/Footer.module.css
```diff
@@ -38,9 +39,20 @@
 
 .subtitle {
   position: absolute;
-  width: max-content;
   top: 4rem;
   left: 5rem;
+  width: max-content;
```
moritzkirstein, line 44: issue: `max-content` for text content results in the problem we where trying to fix and somewhat contradicts `break-word`, see examples at https://developer.mozilla.org/en-US/docs/Web/CSS/max-content

suggestion: we should remove this from the mobile view entirely and only use it for the bigger screen sizes where it does not cause issues

moritzkirstein (changes requested):

moritzkirstein (approved):

PR #624, opened 2024-11-05
fix: footer section & form validation

KevinRohlf:
Fixes #

  - deleted min validation from Parameter Name  
  - rename footer section Privacy to Legal
  - rename link Privacy to Privacy Policy

### DIFF
--- a/content/site.json
+++ b/content/site.json
@@ -169,7 +169,7 @@
   "footer": {
     "subtitle": "Transforming Data Act Compliance into Scalable Monetization Opportunities",
     "copyright": "Built by [deltaDAO](https://delta-dao.com)",
-    "privacyTitle": "Privacy",
+    "privacyTitle": "LEGAL",
     "content": [
       {
         "title": "External",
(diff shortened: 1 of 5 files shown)


content/site.json
```diff
@@ -169,7 +169,7 @@
   "footer": {
     "subtitle": "Transforming Data Act Compliance into Scalable Monetization Opportunities",
     "copyright": "Built by [deltaDAO](https://delta-dao.com)",
-    "privacyTitle": "Privacy",
+    "privacyTitle": "LEGAL",
```
moritzkirstein, line 172: issue(blocking): lets use normal capitilization here when defining content. The style is then defined in css

src/components/@shared/FormInput/InputElement/ConsumerParameters/_validation.ts
```diff
@@ -34,7 +34,7 @@ export const validationConsumerParameters: {
       )
       return occasions.length === 1
     })
-    .min(4, (param) => `Name must be at least ${param.min} characters`)
+    /* .min(1, (param) => `Name must be at least ${param.min} characters`) */
```
moritzkirstein, line 37: question(blocking): why did we remove the line entirely instead of lowering to 1?

moritzkirstein (approved):

PR #632, opened 2024-11-20
feat: add custom filter function

KevinRohlf:
## Proposed Changes

  - add a custom filter function

### DIFF
--- a/filters.config.js
+++ b/filters.config.js
@@ -1,4 +1,38 @@
+const FILTER_VALUES = {
+  MUST_EXIST: 'MUST_EXIST',
+  MUST_EXISTS_AND_NON_EMPTY: 'MUST_EXISTS_AND_NON_EMPTY'
+  //...
+}
+
 module.exports = {
-  filters: [],
+  filters: [
+    {
+      id: 'gaiax',
+      label: 'Gaia-X Service',
+      type: 'filterList',
+      options: [
+        // a new filter value for a MUST_EXIST type could be added to handle this new functionality
+        {
+          label: 'Service SD',
+          value: FILTER_VALUES.MUST_EXISTS_AND_NON_EMPTY,
+          queryPath:
+            'metadata.additionalInformation.gaiaXInformation.serviceSD.url'
+        },
+        {
+          label: 'Terms and Conditions',
+          value: FILTER_VALUES.MUST_EXIST,
+          queryPath:
+            'metadata.additionalInformation.gaiaXInformation.termsAndConditions'
+        },
+        // options can have their own queryPath defined that gets validated against the defined value
+        {
+          label: 'Verified',
+          value: FILTER_VALUES.MUST_EXIST,
+          queryPath:
+            'metadata.additionalInformation.gaiaXInformation.serviceSd.isValid'
+        }
+      ]
+    }
+  ],
   filterSets: {}
 }
--- a/src/@types/aquarius/BaseQueryParams.d.ts
+++ b/src/@types/aquarius/BaseQueryParams.d.ts
@@ -3,6 +3,21 @@ interface EsPaginationOptions {
   size?: number
 }
 
+interface boolFilter {
+  bool: {
+    must: {
+      exists: {
+        field: string
+      }
+    }
+    must_not?: {
+      term: {
+        [key: string]: string
+      }
+    }
+  }
+}
+
 interface BaseQueryParams {
   chainIds: number[]
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
@@ -11,6 +26,7 @@ interface BaseQueryParams {
   sortOptions?: SortOptions
   aggs?: any
   filters?: FilterTerm[]
+  bool?: boolFilter[]
   ignorePurgatory?: boolean
   ignoreState?: boolean
   showSaas?: boolean
--- a/src/@utils/aquarius/index.ts
+++ b/src/@utils/aquarius/index.ts
@@ -41,6 +41,18 @@ export function escapeEsReservedCharacters(value: string): string {
  */
 type TFilterValue = string | number | boolean | number[] | string[]
 type TFilterKey = 'terms' | 'term' | 'match' | 'match_phrase'
+type Query = {
+  must: {
+    exists: {
+      field: string
+    }
+  }
+  must_not?: {
+    term: {
+      [key: string]: string
+    }
+  }
+}
 
 export function getFilterTerm(
   filterField: string,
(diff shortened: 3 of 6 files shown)


src/@types/aquarius/BaseQueryParams.d.ts
```diff
@@ -11,6 +26,7 @@ interface BaseQueryParams {
   sortOptions?: SortOptions
   aggs?: any
   filters?: FilterTerm[]
+  bool?: boolFilter[]
```
Abrom8, line 29: suggestion(blocking): Use a const name like `boolFilter` instead of generic `bool`.

src/@utils/aquarius/index.ts
```diff
@@ -56,6 +68,39 @@ export function getFilterTerm(
...
+  }
+
+  let filter: Query[] = []
+  filters.forEach((filterItem) => {
+    let query: Query = {
+      must: {
+        exists: { field: filterItem[0] }
+      }
+    }
+    if (filterItem[1] === 'MUST_EXISTS_AND_NON_EMPTY') {
```
Abrom8, line 90: suggestion(blocking): Move FILTER_VALUES const into this file and reference it to prevent errors: `FILTER_VALUES.MUST_EXISTS_AND_NON_EMPTY`

filters.config.js
```diff
@@ -1,4 +1,38 @@
+const FILTER_VALUES = {
```
Abrom8, line 1: suggestion: Move the FILTER_VALUES const into `aquarius/index.ts`

filters.config.js
```diff
@@ -1,4 +1,38 @@
...
+
 module.exports = {
-  filters: [],
+  filters: [
+    {
+      id: 'gaiax',
+      label: 'Gaia-X Service',
+      type: 'filterList',
+      options: [
+        // a new filter value for a MUST_EXIST type could be added to handle this new functionality
```
Abrom8, line 14: todo(blocking): This seems to be a leftover. Remove it or enhance for clarity.

filters.config.js
```diff
@@ -1,4 +1,38 @@
...
+          queryPath:
+            'metadata.additionalInformation.gaiaXInformation.serviceSD.url'
+        },
+        {
+          label: 'Terms and Conditions',
+          value: FILTER_VALUES.MUST_EXIST,
+          queryPath:
+            'metadata.additionalInformation.gaiaXInformation.termsAndConditions'
+        },
+        // options can have their own queryPath defined that gets validated against the defined value
```
Abrom8, line 27: todo(blocking): This seems to be a leftover. Remove it or enhance for clarity.

src/@types/aquarius/BaseQueryParams.d.ts
```diff
@@ -3,6 +3,21 @@ interface EsPaginationOptions {
   size?: number
 }
 
+interface boolFilter {
```
Abrom8, line 6: suggestion(blocking): Use PascalCase for interface names: `BoolFilter`

src/@utils/aquarius/index.ts
```diff
@@ -41,6 +41,18 @@ export function escapeEsReservedCharacters(value: string): string {
...
+  must: {
+    exists: {
+      field: string
+    }
+  }
+  must_not?: {
+    term: {
+      [key: string]: string
+    }
+  }
```
Abrom8, line 54: suggestion(blocking): This seems to have a lot of duplication with `interface BoolFilter`. Could be suitable to create another interface which is reused in `interface BoolFilter` and here.

src/@utils/aquarius/index.ts
```diff
@@ -41,6 +41,18 @@ export function escapeEsReservedCharacters(value: string): string {
  */
 type TFilterValue = string | number | boolean | number[] | string[]
 type TFilterKey = 'terms' | 'term' | 'match' | 'match_phrase'
+type Query = {
```
Abrom8, line 44: suggestion: remove in favor of new interface next to `BoolFilter` in `aquarius/index.ts`. Sidenote: `Query` name seems too generic since only one specific subset of query is meant by this.

src/@utils/aquarius/index.ts
```diff
@@ -56,6 +68,39 @@ export function getFilterTerm(
   }
 }
 
+export function getFilter(...args: any[]) {
+  let filters = []
+  if (typeof args[0] === 'object') {
+    args[0].forEach((arg) => {
```
Abrom8, line 74: suggestion(blocking): In general prefer usage of `for...of` over `forEach` for readability and performance. More context: https://biomejs.dev/linter/rules/no-for-each/ 

This comment applies to all places in the code where `forEach` is being used.

Abrom8 (changes requested):

src/@types/aquarius/BaseQueryParams.d.ts
moritzkirstein: this is a requirement from elasticsearch, as they expect this to be named `"bool"` in the request body

filters.config.js
KevinRohlf: I made the const into an enum and added it to the other methods in the searchQuery.ts is this ok?

filters.config.js
```diff
@@ -1,4 +1,55 @@
...
+    },
+    {
+      id: 'accessType',
+      label: 'Access Type',
+      type: 'filterList',
+      options: [
+        { label: 'download', value: FilterByAccessOptions.Download },
+        { label: 'compute', value: FilterByAccessOptions.Compute }
+      ]
+    },
```
moritzkirstein, line 27: suggestion(blocking): to reduce complexity and avoid having to define filters differently depending on the `id` we should simply move these filters over to the new specs. This would then look something like the following:

```ts
{
  id: 'accessType',
  label: 'Access Type',
  type: 'filterList',
  options: [
    {
      label: 'Download',
      value: FilterByAccessOptions.Download,
      queryPath: 'services.type'
    },
    {
      label: 'Compute',
      value: FilterByAccessOptions.Compute,
      queryPath: 'services.type'
    }
    //...
  ]
}        
//...
```

src/components/Search/Filter.tsx
```diff
@@ -114,10 +96,37 @@ export default function Filter({
...
-      ? { ...filters, [filterId]: filters[filterId].filter((e) => e !== value) }
-      : { ...filters, [filterId]: [...filters[filterId], value] }
+  async function handleSelectedFilter(
+    option: { label: string; value: string; queryPath?: string },
+    filterId: string
+  ) {
+    let updatedFilters
+    if (option.queryPath) {
+      updatedFilters = filters[filterId].includes(
+        `${option.queryPath}=${option.value}`
```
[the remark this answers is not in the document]
moritzkirstein: this also will get more readable once adapting existing filters to the new, more flexible structure and a lot of code can be removed here.

moritzkirstein (changes requested):

PR #634, opened 2024-11-25
feat: check credential ID match warning

### LINKED ISSUE #618: [BUG] Credential ID check only checks for the first serviceOffering found
## Summary

Credential ID check on GX credentials only checks until it finds first `serviceOffering` instead of validating the parent serviceOffering is used.

## Current Behavior

If a valid `serviceOffering` credential is contained in the presentation, but not the first one found, the credential id check fails.
  
## Expected Behavior

If a valid `serviceOffering` credential is contained in the presentation, the credential id check succeeds, regardless of its position.

Example here: https://portal.pontus-x.eu/asset/did:op:76a8ac58bf8ae7aa97494f00b565c4d1799b4496395c77f0f84dcf879fb28e56

## Possible Solution
Introduce additional step, steps would then be as follows:
1. Get credential data
2. Create list with all credentials of type `gx:ServiceOffering`
3. Read `gx:dependsOn` field in all serviceOffering credentials
4. Remove credentials with IDs found in dependsOn
5. Check ID against last remaining credential

KevinRohlf:
Fixes #618 

## Proposed Changes

  - compare service offering credential id with gx:dependsOn to filter out the correct ID
  - add new state 'warning' to inform you when gx:dependsOn does not exist

### DIFF
--- a/src/@context/Asset.tsx
+++ b/src/@context/Asset.tsx
@@ -43,7 +43,7 @@ export interface AssetProviderValue {
   serviceCredentialIdMatch?: boolean
   serviceCredentialVersion: string
   verifiedServiceProviderName: string
-  matchVerifiable?: boolean
+  isIdMatchVerifiable?: boolean | string
   fetchAsset: (token?: CancelToken) => Promise<void>
 }
 
@@ -83,7 +83,9 @@ function AssetProvider({
     useState<string>()
   const [verifiedServiceProviderName, setVerifiedServiceProviderName] =
     useState<string>()
-  const [matchVerifiable, setMatchVerifiable] = useState<boolean>()
+  const [isIdMatchVerifiable, setisIdMatchVerifiable] = useState<
+    boolean | string
+  >()
 
   const newCancelToken = useCancelToken()
   const isMounted = useIsMounted()
@@ -204,14 +206,14 @@ function AssetProvider({
           ? await getServiceCredential(serviceCredential?.url)
           : serviceCredential?.raw
 
-        const { verified, complianceApiVersion, idMatch, matchVerifiable } =
+        const { verified, complianceApiVersion, idMatch, isIdMatchVerifiable } =
           await verifyRawServiceCredential(serviceCredentialContent, asset.id)
 
         setIsServiceCredentialVerified(verified && !!serviceCredentialContent)
         setServiceCredentialIdMatch(
-          verified && !!serviceCredentialContent && idMatch && matchVerifiable
+          verified && !!serviceCredentialContent && idMatch
         )
-        setMatchVerifiable(matchVerifiable)
+        setisIdMatchVerifiable(isIdMatchVerifiable)
         setServiceCredentialVersion(complianceApiVersion)
         const serviceProviderName = getPublisherFromServiceCredential(
           serviceCredentialContent
@@ -325,7 +327,7 @@ function AssetProvider({
           serviceCredentialIdMatch,
           serviceCredentialVersion,
           verifiedServiceProviderName,
-          matchVerifiable
+          isIdMatchVerifiable
         } as AssetProviderValue
       }
     >
--- a/src/@context/Asset.tsx
+++ b/src/@context/Asset.tsx
@@ -43,6 +43,7 @@ export interface AssetProviderValue {
   serviceCredentialIdMatch?: boolean
   serviceCredentialVersion: string
   verifiedServiceProviderName: string
+  matchVerifiable?: boolean
   fetchAsset: (token?: CancelToken) => Promise<void>
 }
 
@@ -82,6 +83,7 @@ function AssetProvider({
     useState<string>()
   const [verifiedServiceProviderName, setVerifiedServiceProviderName] =
     useState<string>()
+  const [matchVerifiable, setMatchVerifiable] = useState<boolean>()
 
   const newCancelToken = useCancelToken()
   const isMounted = useIsMounted()
@@ -202,13 +204,14 @@ function AssetProvider({
           ? await getServiceCredential(serviceCredential?.url)
           : serviceCredential?.raw
 
-        const { verified, complianceApiVersion, idMatch } =
+        const { verified, complianceApiVersion, idMatch, matchVerifiable } =
           await verifyRawServiceCredential(serviceCredentialContent, asset.id)
 
         setIsServiceCredentialVerified(verified && !!serviceCredentialContent)
         setServiceCredentialIdMatch(
-          verified && !!serviceCredentialContent && idMatch
+          verified && !!serviceCredentialContent && idMatch && matchVerifiable
         )
+        setMatchVerifiable(matchVerifiable)
         setServiceCredentialVersion(complianceApiVersion)
         const serviceProviderName = getPublisherFromServiceCredential(
           serviceCredentialContent
@@ -321,7 +324,8 @@ function AssetProvider({
           isServiceCredentialVerified,
           serviceCredentialIdMatch,
           serviceCredentialVersion,
-          verifiedServiceProviderName
+          verifiedServiceProviderName,
+          matchVerifiable
         } as AssetProviderValue
       }
     >
--- a/src/@context/Asset.tsx
+++ b/src/@context/Asset.tsx
@@ -83,7 +83,7 @@ function AssetProvider({
     useState<string>()
   const [verifiedServiceProviderName, setVerifiedServiceProviderName] =
     useState<string>()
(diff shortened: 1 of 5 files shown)


src/components/@shared/VerifiedBadge/index.tsx
```diff
@@ -22,11 +25,25 @@ export function Badge({
...
+        isValid ? (
+          <VerifiedPatch />
+        ) : (
+          <Cross />
+        )
+      ) : isValid && matchVerifiable ? (
+        <VerifiedPatch />
+      ) : (
+        <Cross />
+      )}
```
moritzkirstein, line 46: suggestion(blocking): this could be simplified with a `false` check, e.g.
```tsx
{isValid && matchVerifiable !== false ? 
  //...
:
  // ... 
}
```

src/@context/Asset.tsx
```diff
@@ -82,6 +83,7 @@ function AssetProvider({
     useState<string>()
   const [verifiedServiceProviderName, setVerifiedServiceProviderName] =
     useState<string>()
+  const [matchVerifiable, setMatchVerifiable] = useState<boolean>()
```
moritzkirstein, line 86: suggestion(blocking): I would suggest naming this `idMatchVerifiable` or `isIdMatchVerifiable` for easier readability

src/components/Publish/_utils.ts
```diff
@@ -506,16 +507,27 @@ export async function verifyRawServiceCredential(
...
+        .map(
+          (credential) => credential?.credentialSubject['gx:dependsOn'][0].id
+        )
+      const serviceOffering =
+        parsedServiceCredential.verifiableCredential.filter(
+          (credential) =>
+            credential?.credentialSubject?.type === 'gx:ServiceOffering' &&
+            credential?.credentialSubject?.id !== dependsOn?.[0]
+        )
+      const credentialId = serviceOffering[0]?.credentialSubject?.id
```
moritzkirstein, line 524: issue(blocking): with this approach we always use the first (`[0]`) array entry of `dependsOn` and `serviceOffering`. We need to make sure to evaluate all `serviceOfferings` found in the credentials array.

pseudo code would be something like:
```
serviceOfferings = parsedCredential.filter(type = 'serviceOffering')
dependOnIds = serviceOfferings.map(service.dependsOn)
rootService = serviceOfferings.filter(!dependOnIds.includes(service.id))

if(rootService.length !== 1)
  // could not be verified
else
  // match asset id against rootService id
```

moritzkirstein: suggestion: for easier mapping of arrays in this case (e.g. `gx:dependsOn`), have a look at the [`flatMap()` array method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap)

src/components/Publish/_utils.ts
```diff
@@ -506,16 +507,27 @@ export async function verifyRawServiceCredential(
...
 
       return {
         verified: true,
         complianceApiVersion,
-        idMatch: did && did?.toLowerCase() === credentialId?.toLowerCase()
+        idMatch: did && did?.toLowerCase() === credentialId?.toLowerCase(),
+        matchVerifiable: dependsOn.length > 0
```
moritzkirstein, line 530: issue(blocking): the match can be verifiable independent of the `dependsOn` length. What matters is the amount of possible "root services", which in all valid cases should be 1

suggestion: see pseude code above for details

src/components/@shared/VerifiedBadge/index.tsx
```diff
@@ -22,11 +25,25 @@ export function Badge({
       className={cx({
         mainLabel: true,
         isValid,
+        isWarning: matchVerifiable === false,
         [className]: className
       })}
     >
+      {matchVerifiable === false && (
+        <Tooltip content="gx:dependsOn could not be found" />
+      )}
```
moritzkirstein, line 34: suggestion(blocking): it is possible to wrap items in a tooltip component for a bit better UX

right now, only hovering the little icon pops up the tooltip, but it would be useful if the tooltip always shows when hovering the text. This can be part of the component overall, allowing to pass an additional string to display as tooltip


image

src/components/@shared/VerifiedBadge/index.tsx
```diff
@@ -22,11 +25,25 @@ export function Badge({
       className={cx({
         mainLabel: true,
         isValid,
+        isWarning: matchVerifiable === false,
         [className]: className
       })}
     >
+      {matchVerifiable === false && (
+        <Tooltip content="gx:dependsOn could not be found" />
```
moritzkirstein, line 33: issue(blocking): we need to change the text to something that is better communicating why the id could not be matched
`dependsOn` is not a necessary field, so it is possible that it is missing.

moritzkirstein (changes requested):

src/components/@shared/VerifiedBadge/index.tsx
moritzkirstein: comment: you can use some dummy texts for now for the different tooltips

src/components/Publish/_utils.ts
```diff
@@ -485,7 +485,7 @@ export async function verifyRawServiceCredential(
   verified: boolean
   complianceApiVersion?: string
   idMatch?: boolean
-  matchVerifiable?: boolean
+  isIdMatchVerifiable?: boolean | string
```
moritzkirstein, line 488: suggestion(blocking): given the recent refactorings it looks like it could make sense reworking this variable a bit. It no longer needs to serve as a boolean value, but rather is sort of a warning message that should be displayed in case some potential verification issues have been found.

We can simply use an optional `string` value here and evaluate based on its presence if a warning should be displayed to the user.

moritzkirstein (changes requested):

src/components/Publish/_utils.ts
```diff
@@ -506,16 +507,53 @@ export async function verifyRawServiceCredential(
...
+          .map((service) => service?.credentialSubject?.id)
+
+        if (rootService.length > 1) {
+          return {
+            verified: true,
+            complianceApiVersion,
+            idMatch: rootService?.includes(did || did.toLowerCase()),
+            isIdMatchVerifiable: 'Too many root services'
+          }
+        } else if (rootService.length === 0) {
```
moritzkirstein, line 543: suggestion/nit-pick(non-blocking): I'd prefer `.length < 1` for easier readability and understanding of last statement

src/components/Publish/_utils.ts
```diff
@@ -506,16 +507,53 @@ export async function verifyRawServiceCredential(
...
-        idMatch: did && did?.toLowerCase() === credentialId?.toLowerCase()
+      const serviceOfferings =
+        parsedServiceCredential.verifiableCredential.filter(
+          (credential) =>
+            credential?.credentialSubject?.type === 'gx:ServiceOffering'
+        )
+      if (serviceOfferings.length === 1) {
+        return {
+          verified: true,
+          complianceApiVersion,
```
moritzkirstein, line 518: suggestion(blocking): this can be a local variable for the api success case as it is included in every return statement

moritzkirstein (changes requested):

src/components/Publish/_utils.ts
```diff
@@ -535,22 +537,19 @@ export async function verifyRawServiceCredential(
 
         if (rootService.length > 1) {
           return {
-            verified: true,
-            complianceApiVersion,
+            ...verifiedAndComplianceApiVersion,
             idMatch: rootService?.includes(did || did.toLowerCase()),
```
moritzkirstein, line 541: issue(blocking): the statement will not work for lowerCase check, as the logical or is not supported in this way. An alternative could be to move the `||` outside the function parameter:
`rootService.includes(did) || rootService.includes(did.toLowerCase())`

src/components/Publish/_utils.ts
```diff
@@ -535,22 +537,19 @@ export async function verifyRawServiceCredential(
...
+            ...verifiedAndComplianceApiVersion,
             idMatch: false,
             isIdMatchVerifiable: 'No root service found'
           }
         } else {
           return {
-            verified: true,
-            complianceApiVersion,
+            ...verifiedAndComplianceApiVersion,
             idMatch: rootService?.includes(did || did.toLowerCase())
```
moritzkirstein, line 553: issue(blocking): see above

moritzkirstein (changes requested):

moritzkirstein (approved):

PR #651, opened 2025-01-19
fix: update purchase success message based on payment mode

KevinRohlf:
## Proposed Changes

  - update the purchase success message on payment mode Subscription

### DIFF
--- a/src/components/Asset/AssetActions/Download/index.tsx
+++ b/src/components/Asset/AssetActions/Download/index.tsx
@@ -388,7 +388,10 @@ export default function Download({
             <div className={styles.confettiContainer}>
               <SuccessConfetti
                 success={`You successfully bought this ${
-                  asset.metadata.type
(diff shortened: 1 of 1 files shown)


src/components/Asset/AssetActions/Download/index.tsx
```diff
@@ -388,7 +388,10 @@ export default function Download({
             <div className={styles.confettiContainer}>
               <SuccessConfetti
                 success={`You successfully bought this ${
-                  asset.metadata.type
+                  asset?.metadata?.additionalInformation?.saas?.paymentMode ===
```
moritzkirstein, line 391: suggestion(blocking): it would be cleaner to bind this message to the service being a `saas` in general. We should check if the asset is marked as `saas` and if it is display `"service"` over `metadata.type`

moritzkirstein (changes requested):

moritzkirstein (approved):

PR #652, opened 2025-01-19
fix: update tests

KevinRohlf:
Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/src/@utils/aquarius/index.test.ts
+++ b/src/@utils/aquarius/index.test.ts
@@ -15,7 +15,7 @@ const defaultBaseQueryReturn: SearchQuery = {
     bool: {
       filter: [
         { terms: { chainId: [1, 3] } },
-        { term: { _index: 'v510' } },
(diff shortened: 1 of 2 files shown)


src/@utils/aquarius/index.test.ts
```diff
@@ -15,7 +15,7 @@ const defaultBaseQueryReturn: SearchQuery = {
     bool: {
       filter: [
         { terms: { chainId: [1, 3] } },
-        { term: { _index: 'v510' } },
+        { terms: { _index: ['v510', 'v510'] } },
```
moritzkirstein, line 18: suggestion(blocking): We can remove the duplicate value in the array if possible.

KevinRohlf: this is because two chains are selected

moritzkirstein: can you please add a code comment to make sure we understand that in the future

moritzkirstein (approved):

PR #664, opened 2025-02-17
feat: display verified address names in Credentials component

KevinRohlf:
Fixes #

## Proposed Changes

  - display verified address names in Credentials component
  -
  -

### DIFF
--- a/src/@hooks/useAddressConfig.ts
+++ b/src/@hooks/useAddressConfig.ts
@@ -68,9 +68,7 @@ export function useAddressConfig(): UseAddressConfig {
     )
   }
 
-  const getVerifiedAddressName: (address: string) => string | undefined = (
-    address
-  ) => {
+  const getVerifiedAddressName = (address: string): string | undefined => {
     const addressKey = Object.keys(verifiedAddresses).find((key) => {
       return key.toLowerCase() === address.toLowerCase()
     })
--- a/src/@hooks/useAddressConfig.ts
+++ b/src/@hooks/useAddressConfig.ts
@@ -69,9 +69,9 @@ export function useAddressConfig(): UseAddressConfig {
   }
 
   const getVerifiedAddressName = (address: string): string | undefined => {
-    const addressKey = Object.keys(verifiedAddresses).find((key) => {
-      return key.toLowerCase() === address.toLowerCase()
-    })
+    const addressKey = Object.keys(verifiedAddresses).find(
+      (key) => key.toLowerCase() === address.toLowerCase()
+    )
     const addressName = verifiedAddresses[addressKey]
     return addressName
   }
--- a/src/@hooks/useAddressConfig.ts
+++ b/src/@hooks/useAddressConfig.ts
@@ -22,6 +22,7 @@ export interface UseAddressConfig {
   isDDOWhitelisted: (ddo: AssetExtended) => boolean
   hasFeaturedAssets: () => boolean
   isWhitelistEnabled: () => boolean
+  getVerifiedAddressName: (address: string) => string | undefined
 }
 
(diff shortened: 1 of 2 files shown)


src/components/@shared/FormInput/InputElement/Credential/index.tsx
```diff
@@ -37,6 +39,14 @@ export default function Credentials(props: InputProps) {
...
   }
 
+  const showAddressName = (address: string) => {
+    const addressKey = Object.keys(verifiedAddresses).filter((key) => {
+      return key.toLowerCase() === address.toLowerCase()
+    })
+    const addressName = verifiedAddresses[addressKey[0]]
+    return addressName?.length > 0 ? `${address} (${addressName})` : address
+  }
+
```
moritzkirstein, line 49: suggestion(blocking): in the `Account` component we have a simpler implementation that only uses addresses as keys
if something more in depth is required (like with `toLowerCase` we should move this to the `useAddressConfig` hook instead (e.g. adding. a `getVerifiedAddressName` function)

src/components/@shared/FormInput/InputElement/Credential/index.tsx
```diff
@@ -37,6 +39,14 @@ export default function Credentials(props: InputProps) {
     setValue('')
   }
 
+  const showAddressName = (address: string) => {
+    const addressKey = Object.keys(verifiedAddresses).filter((key) => {
+      return key.toLowerCase() === address.toLowerCase()
+    })
```
moritzkirstein, line 45: suggestion(blocking): instead of filtering, you could use the `find` function to only return the first hit for any given address

moritzkirstein (changes requested):

src/@hooks/useAddressConfig.ts
```diff
@@ -67,13 +68,22 @@ export function useAddressConfig(): UseAddressConfig {
     )
   }
 
+  const getVerifiedAddressName = (address: string) => {
+    const addressKey = Object.keys(verifiedAddresses).find((key) => {
+      return key.toLowerCase() === address.toLowerCase()
+    })
+    const addressName = verifiedAddresses[addressKey]
+    return addressName?.length > 0 ? addressName : undefined
```
moritzkirstein, line 76: suggestion: this return statement now be shortened
```suggestion
    return addressName
```

src/@hooks/useAddressConfig.ts
```diff
@@ -67,13 +68,22 @@ export function useAddressConfig(): UseAddressConfig {
     )
   }
 
+  const getVerifiedAddressName = (address: string) => {
```
moritzkirstein, line 71: suggestion: adding a return type is recommended for type safety, e.g. `Promise<string|undefined>`

moritzkirstein (changes requested):

src/@hooks/useAddressConfig.ts
```diff
@@ -68,12 +68,14 @@ export function useAddressConfig(): UseAddressConfig {
     )
   }
 
-  const getVerifiedAddressName = (address: string) => {
+  const getVerifiedAddressName: (address: string) => string | undefined = (
+    address
+  ) => {
     const addressKey = Object.keys(verifiedAddresses).find((key) => {
       return key.toLowerCase() === address.toLowerCase()
     })
```
moritzkirstein, line 76: hint(non-blocking): just in case you'd like the shorter version, this would be the same as:
```ts
const addressKey = Object.keys(verifiedAddresses).find((key) => key.toLowerCase() === address.toLowerCase())
```
Comes down to personal preference imo

moritzkirstein (approved):

PR #665, opened 2025-02-17
feat: add Hide button to Onboarding header

KevinRohlf:
Fixes #

## Proposed Changes

  - add button to hide onboarding guide
  -
  -

### DIFF
--- a/src/components/@shared/Onboarding/Header.module.css
+++ b/src/components/@shared/Onboarding/Header.module.css
@@ -38,7 +38,6 @@
 .hideButtonContainer {
   position: absolute;
   display: flex;
-  justify-content: center;
   background-color: var(--background-content);
   right: 0;
   top: 0;
--- a/src/components/@shared/Onboarding/Header.module.css
+++ b/src/components/@shared/Onboarding/Header.module.css
@@ -31,7 +31,7 @@
   color: var(--brand-white);
 }
 
-.container svg {
+.container > svg {
   display: none;
 }
 
@@ -48,7 +48,7 @@
 }
 
 @media screen and (min-width: 40rem) {
-  .container svg {
+  .container > svg {
     width: 8rem;
     display: inline;
   }
--- a/src/components/@shared/Onboarding/Header.module.css
+++ b/src/components/@shared/Onboarding/Header.module.css
@@ -7,6 +7,7 @@
   color: var(--brand-white);
   border-radius: var(--border-radius) var(--border-radius) 0 0;
   box-shadow: var(--box-template-box-shadow);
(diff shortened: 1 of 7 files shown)


src/components/@shared/Onboarding/Header.module.css
```diff
@@ -30,12 +31,24 @@
...
 
-.container svg {
+.container > svg {
   display: none;
 }
 
+.hideButtonContainer {
+  position: absolute;
+  display: flex;
+  justify-content: center;
```
moritzkirstein, line 41: comment(non-blocking): this might not be needed here, as padding + aspect-ratio should be enough to center the single item

src/components/@shared/Onboarding/index.tsx
```diff
@@ -46,7 +46,8 @@ export default function OnboardingSection(): ReactElement {
   const { balance } = useBalance()
   const web3Provider = useProvider()
   const { chain } = useNetwork()
-  const { onboardingStep, setOnboardingStep } = useUserPreferences()
+  const { onboardingStep, setOnboardingStep, setShowOnboardingModule } =
+    useUserPreferences()
```
moritzkirstein, line 50: suggestion(blocking): rather make use of the `useUserPreferences` hook in the `HideButton` component itself to avoid having to pass the `setShowOnboardingModule` function through multiple child components

src/components/@shared/Onboarding/HideButton.tsx
```diff
@@ -0,0 +1,26 @@
+import { ReactElement } from 'react'
+import Button from '../atoms/Button'
+import styles from './HideButton.module.css'
+import Tooltip from '../atoms/Tooltip'
+import Markdown from '../Markdown'
+import Cross from '@images/x-cross.svg'
+
+export default function HideButton({
+  setShowOnboardingModule,
```
moritzkirstein, line 9: suggestion(blocking): see comment below: rather import this in the component via the hook

src/components/@shared/Onboarding/HideButton.tsx
```diff
@@ -0,0 +1,26 @@
...
+  hintText
```
moritzkirstein, line 10: suggestion(blocking): the same goes for the hint text in this case: you can import it from content directly in this component

moritzkirstein: comment/hint: this wont be possible always ofc, but in general it is a good idea to try and keep components independent and self contained to ensure easier maintainability and re-usability of components

moritzkirstein (changes requested):

PR #673, opened 2025-02-25
refactor : made custom provider url to default provider url

KevinRohlf:
Fixes #

## Proposed Changes

  - rename customProviderUrl to defaultProviderUrl
  - update defaultProviderUrl to a fallback value


moritzkirstein:
I don't think this refactoring is needed. Closing the PR for now, as this potentially breaks expected behaviors in some portal instances.
We should rather make sure the `customProviderUrl` variable is correctly documented (comment for now) to be an override for all provider calls.

PR #674, opened 2025-02-25
fix: adjust asset selection component to enable selection on edit page

KevinRohlf:
Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/src/components/@shared/FormInput/InputElement/AssetSelection/index.tsx
+++ b/src/components/@shared/FormInput/InputElement/AssetSelection/index.tsx
@@ -93,9 +93,9 @@ export default function AssetSelection({
                   checked={selected && asset.did === selected}
                   defaultChecked={asset.checked}
                   disabled={
-                    isOnEditPage
+                    disabled || isOnEditPage
                       ? false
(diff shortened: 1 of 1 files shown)


src/components/@shared/FormInput/InputElement/AssetSelection/index.tsx
```diff
@@ -90,7 +92,11 @@ export default function AssetSelection({
                   {...props}
                   checked={selected && asset.did === selected}
                   defaultChecked={asset.checked}
-                  disabled={disabled || !asset.isAccountIdWhitelisted}
+                  disabled={
+                    isOnEditPage
+                      ? false
+                      : disabled || !asset.isAccountIdWhitelisted
+                  }
```
moritzkirstein, line 99: issue(blocking): we should make sure the `disabled` prop gains priority. When this is set the selection should always be disabled. Only the id whitelisting needs to be overwritten on the /edit page

moritzkirstein (changes requested):

moritzkirstein (approved):

PR #679, opened 2025-02-26
fix: support 'Pay per use' payment mode in asset actions

KevinRohlf:
Fixes #

## Proposed Changes

  - Pay-per-use interface is displayed if `paymentMode` is `‘Pay per use’` or `‘payperuse’`.
  - Access count is always displayed


moritzkirstein:
This should rather be fixed in the assets metadata. Multiple values are confusing and not according to spec

PR #682, opened 2025-03-04
fix: styling from publisher on MetaInfo

KevinRohlf:
Fixes #

## Proposed Changes

  -
  -
  -


moritzkirstein (approved):
lgtm! I noticed that we only show the name once it was verified by the GX credential. However in other places in the portal we use the PX registry to display adress names. Might be something we want to update (and have credential verification badge be the only GX relation)

PR #695, opened 2025-03-31
refactor: use custom wagmi client without WalletConnect or CoinBase

KevinRohlf:
Fixes #

## Proposed Changes

  - the default client of connectkit is no longer used
  - Connecting with WallectConnect and CoinBase is no longer possible


src/@utils/wallet/index.ts
```diff
@@ -24,18 +26,28 @@ export async function getDummySigner(chainId: number): Promise<Signer> {
...
 
+// Wagmi client
+export const wagmiClient = createClient({
+  autoConnect: true,
+  connectors: [
+    new MetaMaskConnector({
+      chains,
+      options: {
+        shimDisconnect: true
+      }
```
moritzkirstein, line 44: question: in the [current docs](https://1.x.wagmi.sh/core/connectors/metaMask#shimdisconnect) it shows that `true` is the default value for `shimDisconnect`. Is this option needed in the version we use?

KevinRohlf: oh yes, I must have overlooked that

moritzkirstein (approved):

PR #698, opened 2025-04-22
Development to Main

KevinRohlf:
## Proposed Changes

  - #674 
  - #685 
  - #695

### DIFF
--- a/src/@utils/wallet/index.ts
+++ b/src/@utils/wallet/index.ts
@@ -7,7 +7,7 @@ import { getOceanConfig } from '../ocean'
 import { getSupportedChains } from './chains'
 import { chainIdsSupported } from '../../../app.config'
 import { MetaMaskConnector } from 'wagmi/connectors/metaMask'
-import { publicProvider } from 'wagmi/providers/public'
+import { jsonRpcProvider } from 'wagmi/providers/jsonRpc'
 
 export async function getDummySigner(chainId: number): Promise<Signer> {
   if (typeof chainId !== 'number') {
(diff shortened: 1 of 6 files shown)


src/@utils/wallet/index.ts
```diff
@@ -24,18 +25,25 @@ export async function getDummySigner(chainId: number): Promise<Signer> {
...
-  getDefaultClient({
-    appName: 'Pontus-X',
-    infuraId: process.env.NEXT_PUBLIC_INFURA_PROJECT_ID,
-    // TODO: mapping between appConfig.chainIdsSupported and wagmi chainId
-    chains: getSupportedChains(chainIdsSupported),
-    walletConnectProjectId: process.env.NEXT_PUBLIC_WALLETCONNECT_PROJECT_ID
-  })
+const { chains, provider, webSocketProvider } = configureChains(
+  supportedChains,
+  [publicProvider()]
```
Abrom8, line 32: question(blocking): Why use public providers? Can we use our own provider?

src/components/@shared/Page/Seo/index.tsx
```diff
@@ -74,6 +74,11 @@ export default function Seo({
           {JSON.stringify(datasetSchema).replace(/</g, '\\u003c')}
         </script>
       )}
+      <script
+        defer
+        data-domain="portal.pontus-x.eu"
+        src="https://plausible.io/js/script.js"
+      ></script>
```
Abrom8, line 81: suggestion(blocking): Check if https://nextjs.org/docs/messages/no-script-tags-in-head-component is applicable to the plausible use case.

KevinRohlf: Yes, it is applicable, but then the script must change the location to the _app component.
#702

Abrom8 (approved):

PR #713, opened 2025-05-08
fix: enhance error message for algorithm checksum mismatch

KevinRohlf:
Fixes #

## Proposed Changes

  - enhance error message for algorithm checksum mismatch

### DIFF
--- a/src/@utils/provider.ts
+++ b/src/@utils/provider.ts
@@ -22,7 +22,7 @@ import { KeyValuePair } from '@shared/FormInput/InputElement/KeyValueInput'
 import { Signer } from 'ethers'
 import { getValidUntilTime } from './compute'
 import { toast } from 'react-toastify'
-import { oceanErrorStates } from 'src/@constants/errors'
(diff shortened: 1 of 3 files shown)


src/@utils/provider.ts
```diff
@@ -56,9 +56,15 @@ export async function initializeProviderForCompute(
       accountId
     )
   } catch (error) {
-    const message = getErrorMessage(error.message)
-    LoggerInstance.error('[Initialize Provider] Error:', message)
-    toast.error(message)
+    const { message } = error
+    if (message.includes('algorithm_file_checksum_mismatch')) {
```
moritzkirstein, line 60: suggestion: use a constant that can be quickly accessed and changed. Could be useful overall, maybe we can think about a "ocean_error_states" constant or something like that.

moritzkirstein (changes requested):

moritzkirstein (changes requested):

moritzkirstein (approved):

PR #718, opened 2025-05-21
feat: implement cookie-based user preferences management in UserPreferencesProvider

KevinRohlf:
Fixes #

## Proposed Changes

  - use cookies for user preferences instead of localstorage

### DIFF
--- a/src/@context/UserPreferences.tsx
+++ b/src/@context/UserPreferences.tsx
@@ -7,9 +7,14 @@ import {
   useEffect
 } from 'react'
 import { LoggerInstance, LogLevel } from '@oceanprotocol/lib'
-import { isBrowser } from '@utils/index'
 import { useMarketMetadata } from './MarketMetadata'
 import { AUTOMATION_MODES } from './Automation/AutomationProvider'
+import {
+  deleteCookie,
+  getCookieValue,
+  SAME_SITE_OPTIONS,
+  setCookie
+} from '@utils/cookies'
 
 interface UserPreferencesValue {
   debug: boolean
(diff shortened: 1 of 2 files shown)


src/@context/UserPreferences.tsx
```diff
@@ -40,97 +45,161 @@ interface UserPreferencesValue {
...
   )
 
   const [onboardingStep, setOnboardingStep] = useState<number>(
-    localStorage?.onboardingStep || 0
+    typeof getCookieValue('onboardingModule') !== 'undefined'
+      ? ~~JSON.parse(getCookieValue('onboardingModule'))?.onboardingStep
+      : 0
   )
+  const twoMonthsFromNow = new Date()
+  twoMonthsFromNow.setMonth(twoMonthsFromNow.getMonth() + 2)
```
moritzkirstein, line 108: suggestion: move this to `@utils/cookies`, similar to the 365 days. Could be some defaults to import when setting cookie length

src/@context/UserPreferences.tsx
```diff
@@ -40,97 +45,161 @@ interface UserPreferencesValue {
...
+      ? JSON.parse(getCookieValue('automationWallet'))?.automationWalletJSON
+      : ''
   )
 
   const [automationWalletMode, setAutomationWalletMode] =
     useState<AUTOMATION_MODES>(
-      localStorage?.automationWalletMode || AUTOMATION_MODES.SIMPLE
+      typeof getCookieValue('automationWallet') !== 'undefined'
+        ? JSON.parse(getCookieValue('automationWallet'))?.automationWalletMode
+        : AUTOMATION_MODES.SIMPLE
```
moritzkirstein, line 92: suggestion(non-blocking): would be nice if this can be configured in app.config per portal

src/@context/UserPreferences.tsx
```diff
@@ -40,97 +45,161 @@ interface UserPreferencesValue {
...
+    (typeof getCookieValue('onboardingModule') !== 'undefined' &&
+      JSON.parse(getCookieValue('onboardingModule'))?.showOnboardingModule ===
+        'true') ??
+      showOnboardingModuleByDefault
   )
 
   const [onboardingStep, setOnboardingStep] = useState<number>(
-    localStorage?.onboardingStep || 0
+    typeof getCookieValue('onboardingModule') !== 'undefined'
+      ? ~~JSON.parse(getCookieValue('onboardingModule'))?.onboardingStep
```
moritzkirstein, line 104: question: why are you using the `~` operator here?
suggestion: in general it is preferred to use operators / expressions that communicate well what they are intended for, in this case it is a bit unclear (at least without a comment) and can probably be solved differently to enhance readability of the code

src/@context/UserPreferences.tsx
```diff
@@ -40,97 +45,161 @@ interface UserPreferencesValue {
...
+      })
+    } else {
+      deleteCookie('allowExternalContent')
+    }
+  }, [allowExternalContent])
+
+  useEffect(() => {
+    if (
+      automationWallet !== '' ||
+      automationWalletMode !== AUTOMATION_MODES.SIMPLE
```
moritzkirstein, line 171: suggestion: see above wrt. app.config

src/@context/UserPreferences.tsx
```diff
@@ -40,97 +45,161 @@ interface UserPreferencesValue {
...
+    ) {
+      const automationWalletCookie = {
+        automationWalletJSON: automationWallet,
+        automationWalletMode
+      }
+      setCookie('automationWallet', JSON.stringify(automationWalletCookie), {
+        expires: null,
```
moritzkirstein, line 178: suggestion: add a comment to explain why this is set to `null`

moritzkirstein: additionally, this could be part of the cookies utils constant so that it can be imported where needed

moritzkirstein (changes requested):
nice work! looks like we should be able to move these into cookies rather quickly 👍

src/@context/UserPreferences.tsx
KevinRohlf: I use the `~` operator to convert the cookie value to a number.
For better readability, I can use `Number()`.

src/@utils/cookies.ts
```diff
@@ -6,14 +6,32 @@ export enum SAME_SITE_OPTIONS {
...
   sameSite: SAME_SITE_OPTIONS.STRICT
 }
 
+export const SESSION_COOKIE_OPTIONS: CookieOptions = {
+  expires: undefined, // undefined means session cookie
+  sameSite: SAME_SITE_OPTIONS.STRICT
+}
+
+export const TWO_MONTHS_COOKIE_OPTIONS: CookieOptions = {
+  expires: addTwoMonths(new Date()),
```
moritzkirstein, line 31: suggestion(non-blocking): it would be sufficient / enough imo to just set the `expires` value to e.g. `60` for the TWO_MONTHS option. Would be a bit easier to read and reduce code complexity.

moritzkirstein (approved):

PR #721, opened 2025-05-27
fix: add optional chaining to balance properties for safer access

rawao:
Fixes #

## Proposed Changes

  -  add optional chaining to balance properties for safer access

### DIFF
--- a/src/components/Header/UserPreferences/Automation/index.tsx
+++ b/src/components/Header/UserPreferences/Automation/index.tsx
@@ -24,7 +24,7 @@ export default function Automation(): ReactElement {
     balance?.approved &&
       setHasZeroERC20TokenBalances(
         Object.keys(balance?.approved)?.filter(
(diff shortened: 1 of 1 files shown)


src/components/Header/UserPreferences/Automation/index.tsx
```diff
@@ -21,13 +21,13 @@ export default function Automation(): ReactElement {
     useState<boolean>(false)
 
   useEffect(() => {
-    balance.approved &&
+    balance?.approved &&
       setHasZeroERC20TokenBalances(
-        Object.keys(balance.approved)?.filter(
-          (token) => Number(balance.approved[token]) <= 0
+        Object.keys(balance?.approved)?.filter(
+          (token) => Number(balance?.approved[token]) <= 0
```
moritzkirstein, line 27: suggestion: since optional chaining on the `balance` object was already added to the above check, it is not necessary here and can be removed.

PR #730, opened 2025-06-05
feat: add SaaS type on publish with separate redirect URL and paymentmode fields

KevinRohlf:
Fixes #

## Proposed Changes

  - add SaaS type on publish
  - SaaS is removed from the Dataset publish flow
  - added separate redirect URL and paymentmode fields on SaaS publish flow

### DIFF
--- a/content/publish/form.json
+++ b/content/publish/form.json
@@ -13,7 +13,7 @@
         "name": "type",
         "label": "Asset Type",
         "type": "boxSelection",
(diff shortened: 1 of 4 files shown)


moritzkirstein (changes requested):
nice! already looks quite straight forward.

general feedback on preview:
I think this section could be removed for saas types (as they cant be compute assets anyways - we can just default to 'access'): 
image

I also would remove the "sample file" for now, as I don't think this applies here.

